PR #4, opened 2022-02-19
Package TuxGuitar

cbix:
This is based on the [tuxguitar package base on AUR](https://aur.archlinux.org/pkgbase/tuxguitar).

TODO:
- [x] pass build
- [x] pass reproducible
- [x] check [java package guidelines](https://wiki.archlinux.org/title/Java_package_guidelines). There are basically 2 options:
  - [x] strictly package to [correct directories](https://wiki.archlinux.org/title/Arch_package_guidelines#Directories), or
  - ~put everything under `/opt/tuxguitar` (popular method for large java packages)~
- [x] check comments on AUR for potential fixes
- [x] testing + fixes
- [ ] figure out more licensing stuff:
  - [ ] source + package gervill.jar dependency (used for sampling; [GPL with classpath exception](https://github.com/bluenote10/gervill/blob/master/LICENSE), could not find official sources)
  - [ ] source + package icedtea-sound.jar dependency (used for PulseAudio output; [also GPL with classpath exception](https://icedtea.classpath.org/download/source/) but only included as a binary .jar in the tuxguitar source...)
  - [x] there is some old VST remote stuff where I don't know what it does
  - otherwise remove from the package and see if we can ship with only alsa + jack + fluidsynth plugins
  - Gervill is also the reason why we have to pin java 11, com.sun.media.sound is an internal API which recent JREs don't allow to be used (an idea would be to create a `tuxguitar-gervill` split package)
  - [ ] the LICENSE file deviates from the original LGPL 2.1 text in that it appends a custom copyright statement; do we have to consider this?
  - [x] remove the sf2 file that is shipped with the source
- [x] use latest SWT version (this fixes lots of Gtk issues for me, especially with dark themes)
- [x] fix `doc` and `help` directory access (not working when opening _Documentation_ or _About_ from the app)
- [x] remove perl script that causes missing dependency warning


packages/mamba/PKGBUILD
```diff
@@ -22,7 +23,7 @@ build() {
 }
 
 package() {
-  depends+=('libasound.so' 'libfluidsynth.so' 'libjack.so' 'liblo.so' 'libsigc++' 'libsmf.so')
+  depends+=('libasound.so' 'libfluidsynth.so' 'libjack.so' 'liblo.so' 'libsigc++' 'libsmf.so' 'gcc-libs')
```
SpotlightKid, line 26: AFAICS, @dvzrv only adds soname dependencies within the  `package` function and puts all other (like `gcc-libs`) in the `depends` in the file header.Maybe he can clarify what the strategy is here.

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,113 @@
...
+# Contributor: Sergej Pupykin <pupykin.s+arch@gmail.com>
+# Contributor: |AhIoRoS| < ahioros@gmail.com >
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+pkgbase=tuxguitar
+pkgname=(tuxguitar tuxguitar-common tuxguitar-gtk2)
+pkgver=1.5.5
+pkgrel=2
+pkgdesc="Multitrack guitar tablature editor and player"
+arch=('x86_64' 'i686' 'armv7h')
```
SpotlightKid, line 14: Should this by `any` because the binary is Java?

SpotlightKid (changes requested):

packages/tuxguitar/PKGBUILD
cbix: The `tuxguitar-common` package includes platform-specific libs using JNI. This was also [reported by namcap](https://ci.cbix.de/osam-cologne/archlinux-proaudio/40/1/7):
```
tuxguitar-common E: ELF file ('usr/share/tuxguitar/lib/libicedtea-sound.so') found in an 'any' package.
tuxguitar-common E: ELF file ('usr/share/tuxguitar/lib/libtuxguitar-alsa-jni.so') found in an 'any' package.
tuxguitar-common E: ELF file ('usr/share/tuxguitar/lib/libtuxguitar-fluidsynth-jni.so') found in an 'any' package.
tuxguitar-common E: ELF file ('usr/share/tuxguitar/lib/libtuxguitar-jack-jni.so') found in an 'any' package.
tuxguitar-common E: ELF file ('usr/share/tuxguitar/lib/libtuxguitar-oss-jni.so') found in an 'any' package.
```

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,113 @@
...
+prepare() {
+  cd tuxguitar-$pkgver-src
+  export MAVEN_OPTS="$MAVEN_OPTS -Duser.home=$srcdir"
+  export JAVA_HOME=/usr/lib/jvm/default
+
+  for file in pom.xml TuxGuitar*/pom.xml build-scripts/*/pom.xml build-scripts/native-modules/*/pom.xml; do
+      sed -i -e 's|<source>1\.6</source>|<source>1\.7</source>|g' \
+             -e 's|<target>1\.6</target>|<target>1\.7</target>|g' "$file"
+
+      # https://maven.apache.org/guides/mini/guide-reproducible-builds.html
```
cbix, line 42: I opened an upstream feature request for this already: https://sourceforge.net/p/tuxguitar/feature-requests/96/

packages/mamba/PKGBUILD
dvzrv: The short answer is: It depends.

The longer answer is:
* Makepkg might detect the wrong package providing a given soname, if e.g. thinking in the context of lib32-foo in [multilib] providing `libfoo.so.1-32` and `foo` in [core], [extra] or [community] providing `libfoo.so.1-64`. 
  This is the case when only using a soname in the initial depends array. An alternative can be to first state the package, then the soname it provides (e.g. `depends=(foo libfoo.so)` in the context of the initial depends declaration) which might allow makepkg to resolve this as well, but may have quirks.
* When adding specific depends for a split package one might want to only have a subset of packages (which may or may not provide a soname) in their respective depends. Think of e.g. a split package `foo` and `foo-docs`, where `foo` depends on something, but `foo-docs` does not.

diovudau:
I just tested building, installing and running tuxguitar.

It took relatively long and created three packages but the gtk2 one is optional?  I didn't install it.
Do we really want to provide the gtk2 version?

All seems good, from a user perspective. It works (tested with jack midi).
I am not sure why the Help->About  is empty though. Maybe that is just upstream.

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+_java_ver=11
+_swt_rel=4.22-202111241800
+_swt_ver=${_swt_rel:0:-13}
+url="https://sourceforge.net/projects/tuxguitar/"
+license=('LGPL')
+depends=('alsa-lib' 'libxtst' 'glibc' "java-runtime-openjdk=${_java_ver}" 'hicolor-icon-theme')
+makedepends=('unzip' 'zip' 'ant' 'jack' 'fluidsynth' "java-environment-openjdk=${_java_ver}" 'maven')
+optdepends=('fluidsynth')
+source=(https://downloads.sourceforge.net/tuxguitar/tuxguitar-$pkgver-src.tar.gz
+        https://archive.eclipse.org/eclipse/downloads/drops4/R-${_swt_rel}/swt-${_swt_ver}-gtk-linux-x86_64.zip
```
SpotlightKid, line 24: Why does this package build swt? Can't we just `makedepends` on [swt](https://aur.archlinux.org/packages/swt)? If version 4.21. is too old, `swt` should be updated (I just flagged out-of-date).

cbix:
The gtk2 version was included in the AUR so I kept it. The only difference is the dependencies and one environment variable in the runner script.

Do you know if it's possible to define a ( gtk2 | gtk3 ) dependency somehow so we could make it a single version? What environments would have gtk2 _instead of_ gtk3 anyway? Cinnamon or other gnome 2 derivatives? Maybe the SWT toolkit nowadays doesn't even care, I'll check later how the Eclipse IDE deals with that.

Anyway my legal questions above are probably fine because we are complying with their classpath exception.

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+    -P platform-linux-$_arch \
+    -Dproject.build.outputTimestamp=$SOURCE_DATE_EPOCH \
+    clean install
+  for _i in build-scripts/{tuxguitar,native-modules/tuxguitar-{alsa,oss,jack,fluidsynth}}-linux-$_arch
+  do (
+    cd $_i
+    mvn \
+      -Dproject.build.outputTimestamp=$SOURCE_DATE_EPOCH \
+      clean install
+  ); done
```
SpotlightKid, line 71: (Opinion, i.e. leave it like it is if you disagree)

A bit too clever for my taste. I'd prefer two sections

```bash
  (
    cd build-scripts/tuxguitar-linux-$_arch
    mvn -Dproject.build.outputTimestamp=$SOURCE_DATE_EPOCH \
      clean install
  )

  for mod in alsa oss jack fluidsynth; do
    (
      cd build-scripts/native-modules/tuxguitar-$mod-linux-$_arch
      mvn -Dproject.build.outputTimestamp=$SOURCE_DATE_EPOCH \
        clean install
    )
  done

```

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+}
+
+package_tuxguitar-common () {
+  # tuxguitar
+  cd tuxguitar-$pkgver-src/build-scripts
+  install -d "$pkgdir"/opt
+  cp -a tuxguitar-linux-$_arch/target/tuxguitar-$pkgver-linux-$_arch "$pkgdir"/opt/tuxguitar
+  cp -a native-modules/tuxguitar-{alsa,oss,jack,fluidsynth}-linux-$_arch/target/build/* \
+    "$pkgdir"/opt/tuxguitar/
+
```
SpotlightKid, line 81: I would avoid using `cp -a` (or `cp -rp`) when installing files. If files have the wrong permissions in the build/source tree, they end up with the wrong permissions in the package too.

If installing just a singe file or a bunch of files from within source directory, use `install -D` to create a destination directory and install (a) file(s) into it in one go. Example:

    install -Dm644 tuxguitar-linux-$_arch/target/tuxguitar-$pkgver-linux-$_arch/dist/* "$pkgdir"/opt/tuxguitar/dist

However, that doesn't work if the source directory contains subdirectories (like the `share` directory here). In that case I would restrict the use of `cp -a`' to that subtree and do a `find <dst-dir> -type f -exec chmod 644 {} \;` afterwards, if there's a possibility of the installed files having the wrong permissions.

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+    mvn \
+      -Dproject.build.outputTimestamp=$SOURCE_DATE_EPOCH \
+      clean install
+  ); done
+}
+
+package_tuxguitar-common () {
+  # tuxguitar
+  cd tuxguitar-$pkgver-src/build-scripts
+  install -d "$pkgdir"/opt
```
SpotlightKid, line 77: Shouldn't we follow the [Java package guidelines](https://wiki.archlinux.org/title/Java_package_guidelines) (section "Java packaging on Arch Linux") regarding the install location?

This might require some extra work adapting the starter shell scripts, though.

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+  cp -a tuxguitar-linux-$_arch/target/tuxguitar-$pkgver-linux-$_arch "$pkgdir"/opt/tuxguitar
+  cp -a native-modules/tuxguitar-{alsa,oss,jack,fluidsynth}-linux-$_arch/target/build/* \
+    "$pkgdir"/opt/tuxguitar/
+
+  # icons
+  cd ..
+  for _i in 16 24 32 48 64 96; do
+    _dir="$pkgdir"/usr/share/icons/hicolor/${_i}x${_i}
+    install -d "$_dir"/{apps,mimetypes}
+    install -m644 TuxGuitar/share/skins/Lavender/icon-${_i}x${_i}.png "$_dir"/apps/tuxguitar.png
```
SpotlightKid, line 87: I would prefer for-variables with descriptive names, e.g. here `size`.

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+    for _m in audio-x-{tuxguitar,gtp,ptb}; do
+      ln -sr "$_dir"/apps/tuxguitar.png "$_dir"/mimetypes/$_m.png
+    done
+  done 
+  
+  # misc
+  cd "${srcdir}/tuxguitar-$pkgver-src/misc"
+  install -Dm644 tuxguitar.xml "$pkgdir"/usr/share/mime/packages/tuxguitar.xml
```
SpotlightKid, line 95: If the source file base name is the same as the one of the target file, you can just use:

    install -Dm644 tuxguitar.xml -t "$pkgdir"/usr/share/mime/packages

Saves repeating the file name.

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+
+package_tuxguitar () {
+  depends=('tuxguitar-common' 'gtk3')
+  replaces=('tuxguitar-gtk3')
+  arch=('any')
+  
+  install -Dm 755 tuxguitar.sh "$pkgdir"/usr/bin/tuxguitar
+}
+
+package_tuxguitar-gtk2 () {
```
SpotlightKid, line 110: I'm not sure why this is a separate package if the only difference is the value of an environment variable set in the starter script. Can you clarify?

The `gtk2` dependency of `tuxguitar-gtk2` could be just an `optdepends` of a single `tuxguitar` package, IMHO.

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+      -Dproject.build.outputTimestamp=$SOURCE_DATE_EPOCH \
+      clean install
+  ); done
+}
+
+package_tuxguitar-common () {
+  # tuxguitar
+  cd tuxguitar-$pkgver-src/build-scripts
+  install -d "$pkgdir"/opt
+  cp -a tuxguitar-linux-$_arch/target/tuxguitar-$pkgver-linux-$_arch "$pkgdir"/opt/tuxguitar
```
SpotlightKid, line 78: The files in the `doc` sub-dir should be installed into `"$pkgdir"/usr/share/doc/${pkgname}`.

packages/tuxguitar/PKGBUILD
cbix: Well, it's currently an AUR package. Ideally, our binary repo should only have dependencies that are installable from the main Arch repos or itself. What I could of course do is add swt to our repo as well.

Then again, Java / Maven projects often pull in lots and lots of build and runtime dependencies ([check the build log](https://ci.cbix.de/osam-cologne/archlinux-proaudio/49/1/5)) which are distributed with the final binary (in case of runtime deps) so including SWT here is not a too bad practice :)

packages/tuxguitar/PKGBUILD
cbix: The article states
> As is standard with Arch Linux packages, if the above standards cannot be adhered to without a serious amount of work, the package should be installed in its preferred manner, with the resulting directory located in `/opt`. This is useful for programs that bundle JREs or include customized versions of dependencies, or do other strange or painful tasks.

It seems TuxGuitar partially relies on its default directory structure with the root passed to the program as a `tuxguitar.home.path` variable. But I could try packaging it more strictly to the Arch defaults, that would probably also remove some of `namcap`s complaints 👍

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,118 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Stephan Springer <buzo+arch@Lini.de>
+# Contributor: Thanos Apostolou <thanosapostolou@outlook.com>
+# Contributor: Alexandre Moine <alexandre@moine.me>
+# Contributor: Sergej Pupykin <pupykin.s+arch@gmail.com>
+# Contributor: |AhIoRoS| < ahioros@gmail.com >
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+pkgbase=tuxguitar
+pkgname=tuxguitar
```
SpotlightKid, line 10: When this is changed to not be a split package anymore, `conflicts=(tuxguitar-common and tuxguitar-gtk)` must be added, otherwise you get errors like these on install, if you had the AUR package installed before:

    tuxguitar: /usr/share/applications/tuxguitar.desktop exists in filesystem (owned by tuxguitar-common)

SpotlightKid:
> fix doc and help directory access (not working when opening Documentation or About from the app)

For the about dialog, `/usr/share/tuxguitar/dist` needs to be added to the `CLASSPATH`.

cbix:
>  have no clue how the TGResourceManager determines the base path, though.

Thanks! That TGResourceManager seems to mostly use the `tuxguitar.share.path` variable defined in the launcher script as the base path. Seems to have a different implementation on Android but that's none of our concern :)

Regarding the documentation, I just added a redirect in 435fe9c72b3bf3d7ec16bb4502cd758dbc85730d.

SpotlightKid:
> [remove included sf2 file due to missing license](https://github.com/osam-cologne/archlinux-proaudio/pull/4/commits/9586cec66d445012073520a00f5c3b7c8de1b02b)

Yeah, this soundfont, which seems to have been around for a long time, definitely has unclear origins of its samples.

http://www.personalcopy.com/sfarkfonts1.htm

packages/mamba/PKGBUILD
SpotlightKid: What happend to the `depends+=` in the latest version of the `PKGBUILD`?

SpotlightKid: Oh, I see, these are now optional dependencies.

I still get lots of `namcap` complaint, but they seem all to be bogus. `java-runtime` is already covered by `java-runtime-openjdk` and the other are optional or transient deps.

The "lacks FULL RELRO" could be fixed some ELF tools, I believe.

```
tuxguitar W: ELF file ('usr/lib/tuxguitar/libicedtea-sound.so') lacks FULL RELRO, check LDFLAGS.
tuxguitar W: ELF file ('usr/lib/tuxguitar/libtuxguitar-alsa-jni.so') lacks FULL RELRO, check LDFLAGS.
tuxguitar W: ELF file ('usr/lib/tuxguitar/libtuxguitar-fluidsynth-jni.so') lacks FULL RELRO, check LDFLAGS.
tuxguitar W: ELF file ('usr/lib/tuxguitar/libtuxguitar-jack-jni.so') lacks FULL RELRO, check LDFLAGS.
tuxguitar E: Dependency java-runtime detected and not included (found class files usr/share/java/tuxguitar/commons-compress.jar, usr/share/java/tuxguitar/gervill.jar, usr/share/java/tuxguitar/icedtea-sound.jar, usr/share/java/tuxguitar/itext-pdf.jar, usr/share/java/tuxguitar/itext-xmlworker.jar, usr/share/java/tuxguitar/swt.jar, usr/share/java/tuxguitar/tuxguitar-alsa.jar, usr/share/java/tuxguitar/tuxguitar-ascii.jar, usr/share/java/tuxguitar/tuxguitar-awt-graphics.jar, usr/share/java/tuxguitar/tuxguitar-browser-ftp.jar, usr/share/java/tuxguitar/tuxguitar-community.jar, usr/share/java/tuxguitar/tuxguitar-compat.jar, usr/share/java/tuxguitar/tuxguitar-converter.jar, usr/share/java/tuxguitar/tuxguitar-editor-utils.jar, usr/share/java/tuxguitar/tuxguitar-fluidsynth.jar, usr/share/java/tuxguitar/tuxguitar-gervill.jar, usr/share/java/tuxguitar/tuxguitar-gm-settings.jar, usr/share/java/tuxguitar/tuxguitar-gm-utils.jar, usr/share/java/tuxguitar/tuxguitar-gpx.jar, usr/share/java/tuxguitar/tuxguitar-gtp-ui.jar, usr/share/java/tuxguitar/tuxguitar-gtp.jar, usr/share/java/tuxguitar/tuxguitar-image.jar, usr/share/java/tuxguitar/tuxguitar-jack-ui.jar, usr/share/java/tuxguitar/tuxguitar-jack.jar, usr/share/java/tuxguitar/tuxguitar-jsa.jar, usr/share/java/tuxguitar/tuxguitar-lib.jar, usr/share/java/tuxguitar/tuxguitar-lilypond-ui.jar, usr/share/java/tuxguitar/tuxguitar-lilypond.jar, usr/share/java/tuxguitar/tuxguitar-midi-ui.jar, usr/share/java/tuxguitar/tuxguitar-midi.jar, usr/share/java/tuxguitar/tuxguitar-musicxml.jar, usr/share/java/tuxguitar/tuxguitar-pdf-ui.jar, usr/share/java/tuxguitar/tuxguitar-pdf.jar, usr/share/java/tuxguitar/tuxguitar-ptb.jar, usr/share/java/tuxguitar/tuxguitar-svg.jar, usr/share/java/tuxguitar/tuxguitar-synth-export.jar, usr/share/java/tuxguitar/tuxguitar-synth-gervill.jar, usr/share/java/tuxguitar/tuxguitar-synth.jar, usr/share/java/tuxguitar/tuxguitar-tef.jar, usr/share/java/tuxguitar/tuxguitar-tuner.jar, usr/share/java/tuxguitar/tuxguitar-ui-toolkit-swt.jar, usr/share/java/tuxguitar/tuxguitar-ui-toolkit.jar, usr/share/java/tuxguitar/tuxguitar.jar)
tuxguitar W: Dependency fluidsynth detected but optional (libraries ['usr/lib/libfluidsynth.so.3'] needed in files ['usr/lib/tuxguitar/libtuxguitar-fluidsynth-jni.so'])
tuxguitar W: Dependency jack2 detected but optional (libraries ['usr/lib/libjack.so.0'] needed in files ['usr/lib/tuxguitar/libtuxguitar-jack-jni.so'])
tuxguitar W: Dependency libpulse detected but optional (libraries ['usr/lib/libpulse.so.0'] needed in files ['usr/lib/tuxguitar/libicedtea-sound.so'])
tuxguitar W: Dependency included and not needed ('libxtst')
tuxguitar W: Dependency glibc included but already satisfied
tuxguitar W: Dependency included and not needed ('java-runtime-openjdk')
tuxguitar W: Dependency included and not needed ('gtk3')
``

Comment (cbix) [reply packages/mamba/PKGBUILD]: @SpotlightKid this comment was on the mamba-related changes which are already in master ;) I'll go over the dependencies again some time later, there's still room for optimization (e.g. gtk3 is not needed during build etc)

Comment (SpotlightKid) [reply packages/mamba/PKGBUILD]: Oops. ;)

File: packages/tuxguitar/PKGBUILD [hunk 13]
```diff
@@ -0,0 +1,127 @@
...
+groups=(pro-audio)
+depends=()
+makedepends=(alsa-lib fluidsynth jack "java-environment-openjdk=${_java_ver}" maven)
+optdepends=('alsa-lib: playback with ALSA'
+            'fluidsynth: alternative synthesizer for playback'
+            'jack: playback and MIDI support with JACK'
+            'pulseaudio: playback with PulseAudio')
+conflicts=(tuxguitar-common tuxguitar-gtk2)
+source=("https://downloads.sourceforge.net/tuxguitar/tuxguitar-${pkgver}-src.tar.gz"
+        "https://archive.eclipse.org/eclipse/downloads/drops4/R-${_swt_rel}/swt-${_swt_ver}-gtk-linux-x86_64.zip"
```
SpotlightKid, line 29: Do we need to use a different swt distribution for i686 and ARM 32-bit?

packages/tuxguitar/PKGBUILD
```diff
@@ -0,0 +1,127 @@
...
+# Contributor: Sergej Pupykin <pupykin.s+arch@gmail.com>
+# Contributor: |AhIoRoS| < ahioros@gmail.com >
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+pkgbase=tuxguitar
+pkgname=tuxguitar
+pkgver=1.5.5
+pkgrel=2
+pkgdesc='Multitrack guitar tablature editor and player'
+arch=(x86_64 i686 armv7h)
```
SpotlightKid, line 14: What about aarch64? Or conversely, why include ARM architectures at all? Are we able to produce ARM binary packages?

packages/tuxguitar/tuxguitar.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+export PATH="/usr/lib/jvm/java-11-openjdk/bin:$PATH"
+CP=/usr/share/tuxguitar/dist
+CP="$CP:/usr/share/doc/tuxguitar"
```
SpotlightKid, line 4: Minor niggle: combine these two lines

SpotlightKid (changes requested):

packages/tuxguitar/PKGBUILD
cbix: TG provides build targets for these three architectures (they were also included in the original AUR version). I could probably write another patch to allow building for aarch64 but I'd prefer to "wait for upstream"  for official support.

In general, I can already build + check aarch64 packages ([mamba builds out of the box](https://ci.cbix.de/osam-cologne/archlinux-proaudio/57/2/3)) so we could easily provide basic 64-bit ARM support in our repo as well for packages that successfully build. Personally ARM is something I'd like to support, even though the experience with Arch is currently not fully comparable. Projects like "use your RasPi with Archlinux ARM as a MIDI synth" are definitely something I can imagine some people to consider 😛

(I still keep a RasPi 2 around somewhere so we could even build armv7h without cross compilation, but it will probably lose relevance rather soon)

cbix: Well ok I just tried building for armv7hl and it failed because the SWT version doesn't exist for that architecture. I'll remove armv7h for now.

packages/tuxguitar/tuxguitar.sh
cbix: done in b6d7aa487dbbb94e8837878bfafeda0f27f20c3f

packages/tuxguitar/PKGBUILD
cbix: The last SWT release for i686 was 4.9, which I just pinned in 5b99e0f7f590ae91f8ac176576070c3b8662bfcd

SpotlightKid (approved):

PR #6, opened 2022-03-01
PKGBUILD formatting config

cbix:
This adds an `.editorconfig` to ensure consistent indentation (2 spaces) and (basic) formatting of PKGBUILDs. Also a script to format all PKGBUILDs using [`shfmt`](https://archlinux.org/packages/community/x86_64/shfmt/)

To check if your editor can recognize this file, see https://editorconfig.org/

Maybe someone knows a more powerful tool for linting/formatting PKGBUILDs with configurable best practices.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,6 +11,3 @@ trim_trailing_whitespace = true
 indent_style = spaces
 indent_size = 2
 max_line_length = 100
-binary_next_line = true
-space_redirects = true
-keep_padding = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# archlinux-proaudio editorconfig
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[PKGBUILD]
+indent_style = spaces
+indent_size = 2
+max_line_length = 100
+binary_next_line = true
+space_redirects = true
+keep_padding = true
--- a/tools/fmt.sh
+++ b/tools/fmt.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+#
+# Format PKGBUILD files
+#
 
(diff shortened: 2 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[PKGBUILD]
+indent_style = spaces
+indent_size = 2
+max_line_length = 100
+binary_next_line = true
+space_redirects = true
+keep_padding = true
```
SpotlightKid, line 16: What are those three settings? They are not in the specification, AFAICS.

tools/fmt.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+cd "${0%/*}/.."
+ROOT="$(pwd)"
+
+shfmt -l -i 2 -sr -kp -w ${*:-packages/*/PKGBUILD}
```
SpotlightKid, line 6: When and by whom (manual/CI) should this be used? Maybe add a short explanation or quickstart section to the readme, including any depenedencies, which need to be installed locally (if any)?

SpotlightKid (changes requested):

.editorconfig
cbix: Hmm, you're right, they're not in the spec. I found them in the [shfmt manpage](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#examples) where an analogous `.editorconfig` is provided 🤔 I'll just remove it to avoid confusion, at least with vim-editorconfig these didn't have any effect.

tools/fmt.sh
cbix: ee8828510fe322f3432229444a3a3a64d4c69a1d

SpotlightKid (approved):
LGTM :+1:

PR #7, opened 2022-03-04
feat: allow to pass PACKAGE env var to pipeline to build a single package

SpotlightKid:
Not very elegant, but it works ;)

### DIFF
--- a/tools/syncdeps-all.sh
+++ b/tools/syncdeps-all.sh
@@ -8,8 +8,17 @@ ROOT="$(pwd)"
 MAKEPKG_ARGS="$*"
 
 cd "$ROOT/packages"
-for PKG in *; do
-	cd $PKG
-	makepkg -sorcef --noconfirm $MAKEPKG_ARGS
-	cd ..
+
+if [[ -n "$PACKAGE" ]]; then
+    echo "Caching dependencies for package $PACKAGE..."
+    PACKAGES="$PACKAGE"
+else
+    echo "Caching dependencies for all packages..."
+    PACKAGES="$(ls -1)"
+fi
+
+for PKG in $PACKAGES; do
+    cd $PKG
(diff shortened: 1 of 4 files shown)


tools/syncdeps-all.sh
```diff
@@ -8,8 +8,17 @@ ROOT="$(pwd)"
...
+    echo "Caching dependencies for package $PACKAGE..."
+    PACKAGES="$PACKAGE"
+else
+    echo "Caching dependencies for all packages..."
+    PACKAGES="$(ls -1)"
+fi
+
+for PKG in $PACKAGES; do
+    cd $PKG
+    makepkg -srcf --verifysource --noconfirm $MAKEPKG_ARGS
```
cbix, line 22: I previously had `-o -e` to prevent downloading sources and building, but I like the idea of downloading the sources in this step. Then however we'll need to store them in a place where they're actually available, i.e. add another shared volume in the `.drone.yml` which is used as the `SRCDEST` variable in the Cache + Build + Repro steps.

cbix (changes requested):

tools/syncdeps-all.sh
cbix: This could also be the drone workdir so the downloaded sources persist locally. Aaand to optimize local builds further there could be a `.cache` dir in the same folder for the pacman cache 🤔

SpotlightKid: `-o` also downloads the sources, doesn't it? It must have something to unpack.

cbix: Hmm you're probably right about that :D

cbix (approved):

tools/syncdeps-all.sh
SpotlightKid: Lets's make another issue for that: #10

PR #12, opened 2022-03-18
feat: add setup for nvchecker

SpotlightKid:
See `nvchecker/README.md` included in this changeset for a description.


diovudau (approved):

cbix (approved):
I was wondering if we only need to keep the old versions json in this repo. The new versions are used temporarily and then written to the old version file anyway, right?

SpotlightKid:
> I was wondering if we only need to keep the old versions json in this repo
Yes, I came to the same conclusion. I'll remove`new_ver.json`.

SpotlightKid:
Done in 39dd027.

PR #14, opened 2022-03-19
feat: add package 'midiomatic'

SpotlightKid:


cbix (approved):
Nice work, thanks! Just left a small formatting related comment.

Also I noticed some lv2 packages in the arch repos would list `lv2-host` as an (opt-)depends, do you think that's something we should adopt?

Examples:
- [padthv1](https://archlinux.org/packages/community/x86_64/padthv1/) provides a standalone jack app + lv2 plugin, but has `lv2-host` as a strict dependency
- [lsp-plugins](https://archlinux.org/packages/community/x86_64/lsp-plugins/) provides jack, lv2 and vst versions and has `jack`, `lv2-host` and `vst-host` as optdepends

SpotlightKid:
> [...] list `lv2-host` as an (opt-)depends, do you think that's something we should adopt?

Seems like a good idea in principle, but since atm there's no guarantee that every potential LV2 host adds `lv2-host' to its `provides` I would only put it in `optdepends`, even for packages, which only contains (an) LV2 plugin(s), and definitely for packages providing several plugin formats or additional stand-alone programs. IMHO, that `padthv1` *requires* `lv2-host` is a bug.

You can do useful things with LV2 plugins with all sorts of LV2 tools, which are not typical hosts (e.g. `lv2file`, `lv2proc`, `jack-audio-tools-lv2` etc.), and some hosts will probably never be in the official repos and thus might not provide `lv2-host`.

cbix:
> some hosts will probably never be in the official repos

Good point given the popularity of commercial DAWs on Linux.

PR #15, opened 2022-03-19
feat: add package 'mclk.lv2'

SpotlightKid:


cbix:
This should probably eventually go into the [`x42-plugins`](https://archlinux.org/packages/community/x86_64/x42-plugins/) package from @dvzrv

cbix (approved):

SpotlightKid:
Upstream would have to add it to the `x42-plugins` repo.

cbix:
Oh, right, I wasn't aware of that repo 👍

PR #17, opened 2022-03-21
feat: add package 'python-rtmidi'

SpotlightKid:


SpotlightKid:
`namcap` seems confused about the dependencies but I believe they are correct. The shared lib for the Python extension links also to some libs provided by `glibc` and `gcc-libs` but these are covered by the dependency on `python`. I think it is safe to assume, that python will always (indirectly) depend on these.

diovudau:
> `namcap` seems confused about the dependencies but I believe they are correct. The shared lib for the Python extension links also to some libs provided by `glibc` and `gcc-libs` but these are covered by the dependency on `python`. I think it is safe to assume, that python will always (indirectly) depend on these.

https://wiki.archlinux.org/title/Arch_package_guidelines#Package_dependencies  says to not rely on transitive dependencies and that all direct dependencies should be listed.

SpotlightKid:
Yes, but:

1. `glibc` and `gcc-libs` are so fundamental, that almost every package would have to list them. Thatswhy even official packages usually don't (unless it's the *only* dependency).
2. `python-rtmidi` will never function without `python`. It is `python` which depends on `glibc` (and `gcc-libs`). If Python would switch to another libc, `python-rtmidi` would not require `glibc` anymore either, because it would have to be recompiled to still work with that Python. So it#s not a dependency of `python-rtmidi`.

diovudau:
> Yes, but:
> 
> 1. `glibc` and `gcc-libs` are so fundamental, that almost every package would have to list them. Thatswhy even official packages usually don't (unless it's the _only_ dependency).
> 2. `python-rtmidi` will never function without `python`. It is `python` which depends on `glibc` (and `gcc-libs`). If Python would switch to another libc, `python-rtmidi` would not require `glibc` anymore either, because it would have to be recompiled / re-linked to still work with that Python. So it's not a dependency of `python-rtmidi`.

Very well. These are proper reasons not do to it.

cbix:
> `namcap` seems confused about the dependencies but I believe they are correct.

Those warning are normal if you have nothing installed when checking the package :)

Could you add a simple nvchecker config for this, just for completeness, or do you think that's a stupid idea given your connections to the developer? 😜😜

SpotlightKid:
> Could you add a simple nvchecker config for this

ofc, just forgot it.

PR #22, opened 2022-03-24
feat: add package 'qjackcapture'

SpotlightKid:


packages/qjackcapture/PKGBUILD
```diff
@@ -0,0 +1,26 @@
...
+
+_project=QJackCapture
+pkgname="${_project,,}"
+pkgver=0.1.2
+pkgrel=1
+pkgdesc='A GUI for easy recording of JACK audio sources using jack_capture'
+arch=(any)
+url='https://github.com/SpotlightKid/qjackcapture'
+license=(GPL)
+depends=(hicolor-icon-theme jack_capture python-pyjacklib python-pyqt5)
```
cbix, line 12:
```suggestion
depends=(hicolor-icon-theme jack_capture python-natsort python-pyjacklib python-pyqt5)
```
was missing so I got this error:
```
  File "/usr/bin/qjackcapture", line 33, in <module>
    sys.exit(load_entry_point('QJackCapture==0.1.2', 'console_scripts', 'qjackcapture')())
  File "/usr/bin/qjackcapture", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.10/importlib/metadata/__init__.py", line 171, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/lib/python3.10/site-packages/qjackcapture/app.py", line 47, in <module>
    from natsort import humansorted
ModuleNotFoundError: No module named 'natsort'
```

cbix (changes requested):
I don't know if any other python modules might be missing from the `depends`, but could you add something like `qjackcapture -h` in a `check()` function so next time upstream adds a new module we can immediately see that.

packages/qjackcapture/PKGBUILD
SpotlightKid: In this case, the dependency was there since the initial release, but apparently my AUR PKGBUILD was wrong. :(

Some kind of smoke test in `check()` might make sense, but just running the program with `-h` seems weird. It wouldn't catch all missing imports in any Python program, since some imports might be deferred or not done when only running the command line. In the case of `qjackcapture` it would work, though, yes. I'll have to consider this a bit.

packages/qjackcapture/PKGBUILD
```diff
@@ -0,0 +1,26 @@
...
+makedepends=(python-setuptools)
+groups=(pro-audio)
+source=("https://files.pythonhosted.org/packages/source/q/$pkgname/$_project-$pkgver.tar.gz")
+sha256sums=('b716cee7917468a8c8589dfb3212ba09bfda43224208c96da25b8243716128bf')
+
+build() {
+  cd $_project-$pkgver
+  make
+}
+
```
cbix, line 22: There you go:
```suggestion

check() {
  cd $_project-$pkgver
  python -m qjackcapture -h > /dev/null
}

```
For me this would at lease fail `makepkg` if `python-natsort` is not installed, and passes otherwise.

This of course is nothing compared to a full test suite, but in general we should keep our own checks simple but effective if the project doesn't provide more extensive testing via `make test` or similar.

cbix (changes requested):

cbix (approved):

PR #23, opened 2022-03-24
improve nvchecker patterns and CI

cbix:
Today there were a few incidents where apparently the CI host could not reach the docker registry and fail the pipeline run. This should not happen as all used images are generally available locally already. For regular cache maintenance I'll eventually setup another solution outside of this repo :)

### DIFF
--- a/nvchecker/archlinux-proaudio.toml
+++ b/nvchecker/archlinux-proaudio.toml
@@ -43,7 +43,7 @@ prefix = "v"
 [tuxguitar]
 source = "regex"
 url = "https://sourceforge.net/projects/tuxguitar/rss?path=/TuxGuitar"
-regex = 'tuxguitar-(.+)-src.tar.gz'
+regex = 'tuxguitar-(\d+\.\d+(?:\.\d+)?)-src.tar.gz'
 encoding = "utf-8"
 
 [ykchorus]
--- a/nvchecker/archlinux-proaudio.toml
+++ b/nvchecker/archlinux-proaudio.toml
@@ -7,23 +7,20 @@ newver = "new_ver.json"
 source = "github"
 github = "brummer10/Mamba"
 use_max_tag = true
-from_pattern = '^v'
-to_pattern = ""
+prefix = "v"
 
-[mclk-lv2]
-# note: dots in section names cause errors
+# note: unquoted dots in section names cause errors
+["mclk.lv2"]
 source = "github"
 github = "x42/mclk.lv2"
 use_max_tag = true
-from_pattern = '^v'
-to_pattern = ""
+prefix = "v"
 
 [midiomatic]
 source = "github"
 github = "SpotlightKid/midiomatic"
 use_max_tag = true
-from_pattern = '^v'
-to_pattern = ""
+prefix = "v"
 
 [python-pyjacklib]
 source = "pypi"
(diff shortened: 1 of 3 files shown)


nvchecker/archlinux-proaudio.toml
```diff
@@ -37,18 +34,16 @@ pypi = "python-rtmidi"
...
-to_pattern = ""
+prefix = "v"
 
 [tuxguitar]
 source = "regex"
-url = "http://www.tuxguitar.com.ar/download.html"
-regex = 'Latest Version (\d+\.\d+(?:\.\d+)?)'
+url = "https://sourceforge.net/projects/tuxguitar/rss?path=/TuxGuitar"
+regex = 'tuxguitar-(.+)-src.tar.gz'
 encoding = "utf-8"
```
SpotlightKid, line 43: Can you elaborate why you think this check is better?

SpotlightKid (changes requested):
You should also rebase on latest `master`.

nvchecker/archlinux-proaudio.toml
cbix: The new URL lists all release files in a machine-readable (and -generated) format (RSS/XML) in contrast to the manually updated website.

By using the `tuxguitar-(.+)-src.tar.gz` regex it's also ensured that the source file in the PKGBUILD actually exists for a new version.

SpotlightKid: The website is updated manually, yes, but so are git tags. IMO a software release is inherently a manual process, because it involves a declaration of intent (i.e. the developer declares that a version is ready and the latest official release (or maybe not, in the case of pre-releases). Determining the latest release version from a list of published files is therefor unreliable. The website, instead, clearly states the POV of upstream what is the latest release. If that is not updated reliable, it's a problem of upstream, which we shouldn't try to fix, IMHO.

As far as I understand it, this will [return the highest version from the list](https://github.com/lilydjwg/nvchecker/blob/c6ed37ada1c8b18cc0d06583f051837b091e6a0e/nvchecker/core.py#L316) of matches for the regex. How does this handle pre-release versions?

Side note: RSS is machine-readable, but this check still uses a regex to parse it ;-/ It would be nice, if nvchecker had a plugin for feed formats (e.g. based on feedparser). Maybe feeds could also be parsed withe the [HTML Parser plugin](https://nvchecker.readthedocs.io/en/latest/usage.html#search-with-an-html-parser)? But, as outlined above, I do not think this is the best strategy in this case, anyway.

cbix: This is currently all matches from the RSS:
```
tuxguitar-1.5.5-src.tar.gz
tuxguitar-1.5.4-src.tar.gz
tuxguitar-1.5.3-src.tar.gz
tuxguitar-1.5.2-src.tar.gz
tuxguitar-1.5.1-src.tar.gz
tuxguitar-1.5-src.tar.gz
tuxguitar-1.4-src.tar.gz
tuxguitar-1.3.1-src.tar.gz
tuxguitar-1.3.2-src.tar.gz
tuxguitar-1.3.0-src.tar.gz
```
but for sake of stability I could use a pattern like `tuxguitar-(\d+(\.\d+)+)-src.tar.gz` so it doesn't match things like `tuxguitar-1.5.6rc-src.tar.gz`...

From my understanding SourceForge also creates the "latest release" button based on the list of files so using the RSS which has at least a predictable structure makes sense.

cbix: ecb666bfe007d07d3b47533eac15b01f7c50a57b

You can also see from the [source tarball build script](https://sourceforge.net/p/tuxguitar/code/HEAD/tree/trunk/build-scripts/tuxguitar-src/build.xml) that they're likely to keep the source filenames like that :)

After all, if nvchecker should for some reason detect a wrong update for any package, there's still a human making the PKGBUILD update manually, so ...

SpotlightKid (approved):

PR #28, opened 2022-03-30
package lpd8editor

diovudau:
The cmake instructions in the readme are not for the release tar and there are more problems. 
This is far from ready to pull, I just wanted to "claim" the software for packaging here :)


packages/lpd8editor/PKGBUILD
```diff
@@ -0,0 +1,31 @@
...
+_projectname=lpd8editor
+pkgname="${_projectname,,}"
+pkgver=0.0.15
+pkgrel=1
+pkgdesc="A Linux editor for the Akai LPD8 pad controller."
+arch=(x86_64 aarch64)
+url="https://github.com/charlesfleche/${_projectname}"
+license=('MIT')
+depends=('alsa-lib' 'qt5-base' 'qt5-tools' 'qt5-svg' 'qt5-translations')
+makedepends=('cmake' 'pkgconf')
```
cbix, line 13: `pkgconf` is already part of `base-devel`, so we don't need to include that as mentioned [here](https://wiki.archlinux.org/title/PKGBUILD#makedepends) :)

SpotlightKid:
Yay! Baby steps ... and so on.

A few suggestions:

* Give the branch name a prefix like `package/` or `pkg/`.
* Name the package in the PR title.
* Give the PR the "package" label (@cbix already did this).
* Assign yourself to the PR until its ready for review.
* Don't hesitate to request a review while still in-progress.
* Don't forget to add the package to `nvchecker/archlinux-proaudio.toml` and `nvchecker/old_ver.json`.

packages/lpd8editor/PKGBUILD
```diff
@@ -0,0 +1,31 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Nils Hilbricht <nils -at- hilbricht -dot- net>
+
+_projectname=lpd8editor
+pkgname="${_projectname,,}"
```
SpotlightKid, line 5: `${foo,,}` is just expanded to an all-lowercase version of the value of `foo`, so here you can just use:

```
pkgname=lpd8editor
```
and use `$pkgname` instead of `${_projectname}` in the rest of the file.

packages/lpd8editor/PKGBUILD
SpotlightKid, line 13: @cbix and I settled on the convention to quote shell array members only if necessary, e.g. `depends=(alsa-lib ...)`.

SpotlightKid (commented):
I only commented on some style issues. I left the obvious not-yet-working things uncommented.

cbix (approved):
Hope you don't mind me taking this over :)

PR #29, opened 2022-03-31
Simple website

### LINKED ISSUE #27: Generate web view of available packages at https://arch.osamc.de/
Generate web view of available packages at https://arch.osamc.de/

cbix:
This builds a simple static site using [Hugo](https://gohugo.io) to be pushed to https://arch.osamc.de/.

A preview can be tested under https://arch.osamc.de/hugo-test/, or run `hugo server` from the root dir. To render the package list it expects the [files databases](https://arch.osamc.de/proaudio/aarch64/proaudio.files.tar.gz) to be extracted to `.tmp/db/{x86_64,aarch64}`.

Closes #27 

## TODO
- [x] move some hard-coded strings to the config (repo URL, github link...)
- [x] add some contact information / footer section
- [x] maybe vendor the css lib
- [x] maybe make the table responsive for really small screens 👀
- [x] do SI file size units make sense?
  - yes, most file browsers use SI units
- [x] fix extra line breaks in Chrome
- [x] ~RSS feed for latest package updates~ will try that later

### DIFF
--- a/web/layouts/partials/package-details.html
+++ b/web/layouts/partials/package-details.html
@@ -25,25 +25,25 @@
   {{ if .groups }}
   <tr>
     <th>Groups:</th>
-    <td>{{ range .groups }}{{ . }}<br/>{{ end }}</td>
+    <td>{{ range sort .groups }}{{ . }}<br/>{{ end }}</td>
   </tr>
   {{ end }}
   {{ if .provides }}
   <tr>
     <th>Provides:</th>
-    <td>{{ range .provides }}{{ . }}<br/>{{ end }}</td>
+    <td>{{ range sort .provides }}{{ . }}<br/>{{ end }}</td>
   </tr>
   {{ end }}
   {{ if .conflicts }}
   <tr>
     <th>Conflicts:</th>
-    <td>{{ range .conflicts }}{{ . }}<br/>{{ end }}</td>
+    <td>{{ range sort .conflicts }}{{ . }}<br/>{{ end }}</td>
   </tr>
   {{ end }}
   {{ if .replaces }}
   <tr>
     <th>Replaces:</th>
-    <td>{{ range .replaces }}{{ . }}<br/>{{ end }}</td>
+    <td>{{ range sort .replaces }}{{ . }}<br/>{{ end }}</td>
   </tr>
   {{ end }}
   <tr>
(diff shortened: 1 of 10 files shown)


SpotlightKid:
Looks very nice!

cbix:
Fully responsive now 🎉 (just hiding the description column on small screens)

Screenshot_20220401-102240_Fennec

web/layouts/partials/package-details.html
```diff
@@ -0,0 +1,111 @@
...
+  {{ if .url }}
+  <tr>
+    <th>Upstream URL:</th>
+    <td><a href="{{ .url }}" target="_blank">{{ .url }}</a></td>
+  </tr>
+  {{ end }}
+  {{ if .groups }}
+  <tr>
+    <th>Groups:</th>
+    <td>{{ range .groups }}{{ . }}<br/>{{ end }}</td>
```
SpotlightKid, line 28: Would it be possible to sort groups alphabetically here?

Also for depends, makedepends, provides, conflicts, replces, etc.

cbix: done: 4b6845b5d2e46b5a917a2b8c060feef9cfd689d0

cbix: (the baseURL doesn't fully work with the /hugo-test path but it will work after merging)

SpotlightKid:
The first line of `README.md`

    PKGBUILD files for the binary archlinux pro-audio OSAMC repository.

doesn't really fit for the web page.

cbix:
> The first line of `README.md` doesn't really fit for the web page.

@SpotlightKid I just changed that to

> Actively maintained binary package repo for Arch Linux of free and open source pro-audio software.

Feel free to modify that yourself any time :)

diovudau:
I think for a "nice website" it would be enough to show one architecture on the front page. x86_64 in our case.  It is nice to know that there are other architectures, but the main point is to present what software was packaged. So the lastest version of x64 should be enough.

cbix:
> I think for a "nice website" it would be enough to show one architecture on the front page. x86_64 in our case. It is nice to know that there are other architectures, but the main point is to present what software was packaged. So the lastest version of x64 should be enough.

Hmm, I could make aarch64 another collapsible section so you would click on package → shows details for `x86_64` version + link to expand the aarch64 section of the same package, without listing that twice in the list.

cbix:
@diovudau ok I just created a simple toggle checkbox which is off by default, 815081ecb338231ba8a30e79042cc0a8df0d4c18

diovudau (approved):

SpotlightKid (commented):
We should watch the size of page. Currently it is ~550kB, ~50kB gzipped, which is ok. But when we have lots of packages with lots of files, it could get to obig.

PR #31, opened 2022-04-01
feat: add package 'stone-phaser'

SpotlightKid:


packages/stone-phaser/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+# Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
+
+pkgname=stone-phaser
+pkgver=0.1.2
+pkgrel=1
+pkgdesc='A classic analog phaser effect LADSPA, LV2 and VST2 plugin'
+arch=(x86_64 aarch64)
+url='https://github.com/jpcima/stone-phaser'
+license=(Boost)
+depends=(cairo glibc gcc-libs)
```
cbix, line 11: `glibc` and `gcc-libs` are both pulled in by the `base` package, i.e. part of every Arch installation by default.
The Arch wiki about this:
> and may or may not be listed, for example [glibc](https://archlinux.org/packages/?name=glibc) cannot be uninstalled as every system needs some C library

Since there are no clear rules we could probably decide on our own rules, what do you think?

cbix (changes requested):
Great work!

I sometimes think if those Faust-based plugins were built from the actual Faust sources (instead of shipping transpiled code), they would later benefit from improvements in the Faust compiler itself, maybe. But I agree this way is more reproducible and projects like guitarix do a lot of manual tweaking of the transpiled source...

packages/stone-phaser/PKGBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
+
+pkgname=stone-phaser
+pkgver=0.1.2
+pkgrel=1
+pkgdesc='A classic analog phaser effect LADSPA, LV2 and VST2 plugin'
```
cbix, line 7: Nit: we're sometimes referring to VST2 as just VST (see [midiomatic](https://arch.osamc.de/#x86_64/midiomatic)) and sometimes VST2. Personally, VST2 makes more sense as it's more explicit but then the package group is also called `vst-plugins` (probably for historic reasons)..

cbix (changes requested):

packages/stone-phaser/PKGBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
+
+pkgname=stone-phaser
+pkgver=0.1.2
+pkgrel=1
```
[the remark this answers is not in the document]
SpotlightKid: I think this hinges on a general decision on how we treat the AUR. What actually happens, if the AUR package gets bumped (either pkgrel or version) after the user install our version? (unlikely in this case, since I maintain the AUR package too.)

How would we treat packages where we don't have control over the AUR counterpart?

packages/stone-phaser/PKGBUILD
SpotlightKid: I put these in, since the non-GUI libs do not need cairo, but glibc and gcc-libs are a special case, as you mention.

From observation, he practice in the official repose leans towards including these.

Maybe @dvzrv can give some input?

packages/stone-phaser/PKGBUILD
SpotlightKid: Yes, I think we should distinguish between VST2 and VST3 at least in the description.

The issue with the group name is sth that bugs me too, we should also discuss this with @dvzrv, Either the groud should be renamed to `vst2-plugins` or it should be documented somewhere that `vst-plugins` refers to VST2 and VST3 plugin packges should use `vst3-plugins`.

packages/stone-phaser/PKGBUILD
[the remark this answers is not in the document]
cbix: Most AUR helpers will just pick the highest version number and if repo > AUR, install the repo version.
However if we have a AUR > repo, it's nothing we can control, but since updating an AUR package (again, using an AUR helper like yay) requries manual confirmation and takes longer, most users may not choose to do that.

Anyway this only affects users updating their system with `yay -Syu` and they should know what they're doing.

But let's do some research eventually how other binary repos are dealing with that.

packages/stone-phaser/PKGBUILD
cbix: I see. I'm fine with either solution :)

cbix (approved):

packages/stone-phaser/PKGBUILD
dvzrv: There is unfortunately no (wiki) documentation on the group names (yet).
vst-plugins refers to VST2 and vst3-plugins refers to VST3 plugins though.

A rename of vst-plugins to vst2-plugins is possible but requires to touch/rebuild a few packages.

packages/stone-phaser/PKGBUILD
dvzrv: I always add glibc and gcc-libs to not have transitive dependencies (although every system should have the base package installed, which pulls these two in).

packages/stone-phaser/PKGBUILD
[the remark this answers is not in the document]
SpotlightKid: Bumped pkgrel in ed5b781.

packages/stone-phaser/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+groups=(ladspa-plugins lv2-plugins pro-audio vst-plugins)
+_dpf_commit="05d91f5852f4bccfd2bce1d4d2e2b3036e29db03"
+source=("https://github.com/jpcima/stone-phaser/releases/download/v${pkgver}/${pkgname}-${pkgver}.tar.gz"
+        "${pkgname}-Makefile.base.mk::https://raw.githubusercontent.com/DISTRHO/DPF/$_dpf_commit/Makefile.base.mk")
+sha256sums=('cd19c588469125e4aff3add2a570b4b2d9dc98ef37ccc2785636a0b52df86ddf'
+            'ef890c65422c8c73ffea7687a45317689378ff2b0f3ccadd497ca1cdc77e08a4')
+
+prepare() {
+  cd $pkgname-$pkgver/dpf
+  patch -N -r - -p 1 -i ../resources/patch/DPF-bypass.patch
```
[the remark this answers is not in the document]
SpotlightKid: https://github.com/jpcima/stone-phaser/pull/5

PR #33, opened 2022-04-02
fix: pkgdesc VST -> VST2 and no parallel build

SpotlightKid:
Also fix `build` function to use no parallel make.

### DIFF
--- a/packages/midiomatic/PKGBUILD
+++ b/packages/midiomatic/PKGBUILD
@@ -22,7 +22,7 @@ sha256sums=('aed203bde8fd01e279e058d48f084754f50b9fd758d7b56b845f11d8fbdbb1f3')
 
 build() {
   cd "$pkgname-$pkgver"
-  make
+  make -j1
 }
 
 check() {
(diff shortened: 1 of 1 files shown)


packages/midiomatic/PKGBUILD
```diff
@@ -22,7 +22,7 @@ sha256sums=('aed203bde8fd01e279e058d48f084754f50b9fd758d7b56b845f11d8fbdbb1f3')
 
 build() {
   cd "$pkgname-$pkgver"
-  make
+  make -j1
```
cbix, line 25: Interesting that the `MAKEFLAGS` cause an issue here, but maybe something I should have expected. But I do like this approach, use `-j4` by default and only add `-j1` for packages that don't support parallel builds.

SpotlightKid: This is an issue with midiomatic, which I will fix in the next release.

But it hightlights an issue with our CI. We probably need a job that rebuilds all packages regularly to make sure the tooling didn't break the build process of any packages.

PR #35, opened 2022-04-02
feat: add readme and script for local build/preview for website

SpotlightKid:


cbix (approved):
For development I'd still just run `hugo server` but this could be useful for review :)

SpotlightKid:
> For development I'd still just run `hugo server` 

Yes, but this was only documented in the PR. :)

PR #40, opened 2022-04-03
Package vcf-lv2

cbix:
https://github.com/linuxmao-org/vcf-lv2

### DIFF
--- a/packages/vcf-lv2/PKGBUILD
+++ b/packages/vcf-lv2/PKGBUILD
@@ -25,8 +25,10 @@ check() {
   cd $pkgname-$pkgver
   _lv2_uri='http://jwm-art.net/lv2/vcf'
   for _plugin in plugins/*.lv2; do
-    lv2lint -Mpack -I $_plugin $_lv2_uri/${_plugin:8:-8}{,_cv} ||
-      echo "Ignoring lv2lint errors (known to fail)."
+    lv2lint -Mpack -I $_plugin \
+      -t "Plugin Version *" \
+      -t "Plugin Features" \
+      $_lv2_uri/${_plugin:8:-8}{,_cv}
   done
 }
 
--- a/packages/vcf-lv2/PKGBUILD
+++ b/packages/vcf-lv2/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+pkgname=vcf-lv2
+pkgver=0.0.2
+pkgrel=1
+pkgdesc='VCF LV2 plugins'
+arch=(x86_64 aarch64)
+url='https://github.com/linuxmao-org/vcf-lv2'
+license=(GPL2)
+groups=(lv2-plugins pro-audio)
+depends=(glibc)
+makedepends=(lv2)
+checkdepends=(lv2lint)
+optdepends=('lv2-host: for running the plugins')
+source=("$pkgname-$pkgver.tar.gz::https://github.com/linuxmao-org/$pkgname/archive/refs/tags/v$pkgver.tar.gz")
+sha256sums=('4dc4c7eb3a81cfcbffaabe2c0bac229e6806f984140c6a0f4d0e74c988600bd7')
+
+build() {
+  cd $pkgname-$pkgver
+  make
+}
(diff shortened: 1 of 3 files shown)


packages/vcf-lv2/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+build() {
+  cd $pkgname-$pkgver
+  make
+}
+
+check() {
+  cd $pkgname-$pkgver
+  _lv2_uri='http://jwm-art.net/lv2/vcf'
+  for _plugin in plugins/*.lv2; do
+    lv2lint -Mpack -I $_plugin $_lv2_uri/${_plugin:8:-8}{,_cv} ||
```
SpotlightKid, line 28: This outputs these errors for each plugin:

```
    [FAIL]  Plugin Version Minor
              lv2:minorVersion not found
              seeAlso: <http://lv2plug.in/ns/lv2core#minorVersion>
    [FAIL]  Plugin Version Micro
              lv2:microVersion not found
              seeAlso: <http://lv2plug.in/ns/lv2core#microVersion>
    [FAIL]  Plugin Features
              lv2:[optional|required]Feature <http://lv2plug.in/ns/lv2core#hardRtCapable> not valid
              seeAlso: <http://lv2plug.in/ns/lv2core#Feature>
```

You could ignore these with the `-t ` option, then you don't need to ignore the exit code:

```
  for _plugin in plugins/*.lv2; do
    lv2lint -Mpack -t "Plugin Version *" -t "Plugin Features" \
      -I $_plugin $_lv2_uri/${_plugin:8:-8}{,_cv}
  done
```

cbix: 0875dcc9b5c991a67070b9950ad808f258035b5c

SpotlightKid (commented):
More of a general comment:

We will now have two packages, which only contain LV2 plugins: `mclk.lv2` and `vcf-lv2`. The first uses a "dot l2" suffix, the other "dash lv2". In both cases it's because the upstream is repo is named the way it is.

In the community repo a `-lv2` suffix has sometimes been changed to `.lv2` (example: [mda.lv2](https://archlinux.org/packages/community/x86_64/mda.lv2/)). But it [also has packages](https://archlinux.org/packages/?sort=&q=-lv2) with a `-lv2` suffix. What stance do we want to take here?

cbix:
Good point, I was also thinking about naming it vcf.lv2 but thought the original repo name works well as a `pkgname`.

Regarding a convention of _always_ naming LV2-only packages `*.lv2`:
**Pros**
- consistency
- easier to match LV2-only packages
- some LV2 related tools may be called `something-lv2` so there would be a clear difference between plugin and other packages

**Cons**
- may break with a project's name so user's searching for `vcf-lv2` using pacman may not find it (?)
- `.lv2` sort of looks like a file extension, suggesting a directory called `xyz.lv2` contains an audio plugin. If however `pkgname=xyz.lv2`, there may be directories under `/usr/share/{doc,license}` etc. called `xyz.lv2` _not_ containing an audio plugin. This is not a technical problem at all but maybe a cheap LV2 scanner script would look for dirs named `*.lv2`, idk...
- here, `vcf.lv2` is not actually the name of the resulting plugin (unlike MDA)

SpotlightKid:
> .lv2 sort of looks like a file extension

But that also applies to packages, where upstream is already called `foo.lv2`.

> here, vcf.lv2 is not actually the name of the resulting plugin (unlike MDA)

To be precise: the LV2 bundle dir name. `mda.lv2` actually does not contain an actual plugin called `mda` only the LV2 bundle dir is called `mda.lv2` and the `.lv2` suffix is a required thing. The bundle dir base name is actually of little importance for the user.

SpotlightKid:
This sort convinced my that the renaming of `mda-lv2` to `mda.lv2` was a mistake and that we should stick to the upstream name and don't muck about with unless absolutely necessary. How to determine that name is another question. Sometimes upstream projects are not very clear about that. Is it the repo name, the base-name of the source archive? The title of the readme?

Anyway, in this case, I'd say, let's stick with `vcf-lv2`.

SpotlightKid (approved):

cbix:
Also noticed we're packaging a tool called `jalv.select` from it's repo called `jalv_select` under the package name `jalv-select`. There can't possibly be a single rule to do naming right :D

PR #41, opened 2022-04-04
Package Kapitonov Plugins Pack

cbix:
[**Kapitonov Plugins Pack**](https://github.com/olegkapitonov/Kapitonov-Plugins-Pack) (LADSPA & LV2)

## TODO
- [ ] TBD: `kpp` (still free ;)) or `kapitonov-plugins-pack`?
  - maybe the `ssr` situation is a good example why more explicit = more better?
- [ ] make GUI runtime deps optional?
- [x] how to `check()` this? After `build()`, we only have the individual `.so`s but no ready to test LV2 bundles yet (those are put together by `meson install`)
  - `DESTDIR="test" meson install`
  - lots of lv2lint checks fail, including _Plugin Instantiation_. I believe there is no point now in adding the check there with exceptions as it will not help if in the future the plugins _actually_ break...
- [x] fix namcap errors
- [x] ~package [kpp-vst3](https://github.com/olegkapitonov/KPP-VST3)~ (requires vst3sdk)

### DIFF
--- a/packages/kpp/PKGBUILD
+++ b/packages/kpp/PKGBUILD
@@ -1,0 +1,35 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+_projectname='Kapitonov-Plugins-Pack'
+pkgname=kpp
+pkgver=1.2.1
+pkgrel=1
+pkgdesc='Kapitonov Plugins Pack for guitar sound processing'
+arch=(x86_64 aarch64)
+url='https://kpp-tubeamp.com/'
+license=(GPL3)
+groups=(ladspa-plugins lv2-plugins pro-audio)
+depends=(cairo gcc-libs xcb-util xcb-util-wm)
+makedepends=(boost faust ladspa lv2 meson libxcb zita-convolver zita-resampler)
+optdepends=('ladspa-host: for running the LADSPA plugins'
+            'lv2-host: for running the LV2 plugins')
+source=("$pkgname-$pkgver.tar.gz::https://github.com/olegkapitonov/$_projectname/archive/refs/tags/$pkgver.tar.gz")
+sha256sums=('056c152ee72c5a0cfd45ca0cd848ff13b098c8f58b14c26ad06d7ef49c792b40')
+
+build() {
+  cd $_projectname-$pkgver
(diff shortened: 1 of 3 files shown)


packages/kpp/PKGBUILD
```diff
@@ -0,0 +1,35 @@
...
+makedepends=(boost faust ladspa lv2 meson libxcb zita-convolver zita-resampler)
+optdepends=('ladspa-host: for running the LADSPA plugins'
+            'lv2-host: for running the LV2 plugins')
+source=("$pkgname-$pkgver.tar.gz::https://github.com/olegkapitonov/$_projectname/archive/refs/tags/$pkgver.tar.gz")
+sha256sums=('056c152ee72c5a0cfd45ca0cd848ff13b098c8f58b14c26ad06d7ef49c792b40')
+
+build() {
+  cd $_projectname-$pkgver
+  # default flags are 'auto' so build wouldn't fail if dependencies are missing
+  meson build \
```
SpotlightKid, line 23: You could probably just use `arch-meson` instead.

packages/kpp/PKGBUILD
```diff
@@ -0,0 +1,35 @@
...
+    --prefix /usr \
+    -Dladspa=enabled \
+    -Dlv2=enabled \
+    -Dgui=enabled
+  ninja -C build
```
SpotlightKid, line 28: Just a matter taste: `meson compile`?

cbix: Yea, looks better :) f616b8de1c72fe1a82495e73aea2dd149f0f0014

SpotlightKid (approved):

diovudau:
So it is just  kpp now? I think that is too short. kapitonov-plugins-pack  is a clearer name.

cbix:
> So it is just kpp now? I think that is too short. kapitonov-plugins-pack is a clearer name.

ok I opened #43 for that now. I don't really have a strong opinion, just couldn't find any other software project on the internet that could be called KPP...

SpotlightKid:
I thought about that, but I like short names. And since "kapitonov plugins pack" is in the description, it can be easily found with `pacman` and an AUR website search. Also kpp is easier to remember. For example I always remembered this as "kapit*a*nov plugins". Had I search for that, I wouldn't have found it.

diovudau:
The package name shows up in different places.  pacman -Syu will just show "kpp" in the list.  Maybe it is just me but I use a system update to review if I have packages installed I don't need anymore.

PR #42, opened 2022-04-04
Local drone cache for pacman and sources

### LINKED ISSUE #10: Re-use upstream sources downloaded in "Analyze build scripts" CI step
Quoting @cbix from https://github.com/osam-cologne/archlinux-proaudio/pull/7#pullrequestreview-901118949:

>  like the idea of downloading the sources in this step. Then however we'll need to store them in a place where they're actually available, i.e. add another shared volume in the .drone.yml which is used as the SRCDEST variable in the Cache + Build + Repro steps.

> This could also be the drone workdir (→ just deleting the SRCDIR variable everywhere should be sufficient) so the downloaded sources persist locally. Aaand to optimize local builds further there could be a .cache dir in the same folder for the pacman cache thinking

cbix:
Closes #10 

When running the CI pipeline locally, there is now a `.tmp/cache` directory that persists subsequent runs and caches
- pacman packages
- pacman db
- downloaded sources

This should improve build time when the drone pipeline is used locally.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -42,6 +42,7 @@ steps:
       - name: pkgdb
         path: /var/lib/pacman/sync
     environment:
+      SRCDEST: /drone/src/.tmp/cache/srcdest
       PKGDEST: /drone/src/out
       MAKEFLAGS: -j4
     commands:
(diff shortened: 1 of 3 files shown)


SpotlightKid (approved):
LGTM :+1:

PR #44, opened 2022-04-05
feat: add package 'abcm2ps'

SpotlightKid:


packages/abcm2ps/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+pkgver=8.14.13
+pkgrel=2
+pkgdesc='Convert ABC music notation files to PostScript from the commandline'
+arch=(x86_64 aarch64)
+url='http://moinejf.free.fr/'
+license=(GPL3)
+depends=(pango)
+makedepends=(python-docutils)
+conflicts=(abcm2ps-devel)
+groups=(abc pro-audio)
```
SpotlightKid, line 20: The "abc" group was in the AUR `PKGBUILD`. I can't remember whether I added it. Should I leave it in? (The idea being, that there will be other ABC-related software, like `abc2svg`, `abcmidi` etc., in the same group.)

packages/abcm2ps/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+pkgname=abcm2ps
+pkgver=8.14.13
+pkgrel=2
+pkgdesc='Convert ABC music notation files to PostScript from the commandline'
+arch=(x86_64 aarch64)
+url='http://moinejf.free.fr/'
+license=(GPL3)
+depends=(pango)
+makedepends=(python-docutils)
+conflicts=(abcm2ps-devel)
```
SpotlightKid, line 19: This is also taken from the AUR? `PKGBUILD. I'm not sure when a package name `abcm2ps-devel` ever existed. And if it did, it would be the responsibility of that package to declare the conflict to the release package, IMHO. Leave it in or not?

diovudau: abcm2ps-devel sounds like a debian package. Maybe a converted package? 
If abcm2ps-devel does not exist in AUR I would consider it an error, and remove it.

packages/abcm2ps/PKGBUILD
cbix: Sure, good idea! Probably a few other plugin bundles split up their source in separate projects with separate releases, so it's good practice to group them.

packages/abcm2ps/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+# Contributor: Luis Pablo Gasparotto <lpgasparotto(at)gmail(dot)com>
+
+pkgname=abcm2ps
+pkgver=8.14.13
+pkgrel=2
+pkgdesc='Convert ABC music notation files to PostScript from the commandline'
+arch=(x86_64 aarch64)
+url='http://moinejf.free.fr/'
+license=(GPL3)
+depends=(pango)
```
SpotlightKid, line 17: `abcm2ps` links these libraries:

    -lpangocairo-1.0 -lcairo -lpangoft2-1.0 -lpango-1.0 -lgobject-2.0 -lglib-2.0 -lharfbuzz -lfontconfig -lfreetype -lm

But these are all indirect dependencies of `pango'. I'm still a bit confused about the "declare *transient dependencies*" rule. How do I detect, whether `abcm2ps` needs e.g. `glib2` directly, or whether it is only needed be one (or several)  of the other libraries? Or asked another way, if pango stopped depending (indirectly) on `glib2` and thus wouldn't install it, would `abcm2ps` still work?

packages/abcm2ps/PKGBUILD
SpotlightKid: It existed at some point: https://github.com/aur-archive/abcm2ps-devel

packages/abcm2ps/PKGBUILD
cbix: The [configure script uses pkg-config](https://github.com/lewdlime/abcm2ps/blob/master/configure#L66):
```
CPPFLAGS="$CPPFLAGS -DHAVE_PANGO=1 `$PKG_CONFIG pango cairo freetype2 --cflags`"
LDLIBS="`$PKG_CONFIG pangocairo pangoft2 freetype2 --libs` $LDLIBS"
```
Based on this I'd just list pango, cairo and freetype as dependencies.

diovudau: If pango stopped depending on glib2 and abcm2ps actually needed the glib2 itself our containerized build system will give an error, right? At least that is what I remember from the planning talks with @dvzrv

SpotlightKid: @diovudau This is about run-time, i.e. if the user (hypothetically) installed a pang package, which doesn't require glib2 anymore and glib2 was marked as only installed as a dependency and thus would get de-installed, would abcm2ps still work

TBH, I find this "transitive dependencies" thing highly theoretical anyway. IMHO, in practice this changing of indirect dependencies rarely happens, and, as you say, it will be usually could at build time. Things were easier when the rule was "only include dependencies not already covered by other dependencies", *grumble*.

dvzrv: FWIW: You can use [find-libdeps](https://man.archlinux.org/man/find-libdeps.1) to identify all shared objects directly linked against.

SpotlightKid: > Based on this I'd just list pango, cairo and freetype as dependencies.

Well, you see, based on the output of 

> FWIW: You can use [find-libdeps](https://man.archlinux.org/man/find-libdeps.1) to identify all shared objects directly linked against.

```
libpangocairo-1.0.so=0-64
libpangoft2-1.0.so=0-64
libpango-1.0.so=0-64
libgobject-2.0.so=0-64
libglib-2.0.so=0-64
libfreetype.so=6-64
libm.so=6-64
libc.so=6-64
```

`cairo` is not among the packages which provides any of the these libraries. Do you see why I find this confusing? ;)

Based on the above, I would now arrive at:

```
depends=(glibc)
makedepends=(freetype2 glib2 pango python-docutils)
```

and in `package()`:

```
  depends+=(libpangocairo-1.0.so libpangoft2-1.0.so libpango-1.0.so
            libgobject-2.0.so libglib-2.0.so libfreetype.so)
```

and `cairo` would be supplied by the make dependency on `pango`. The direct mentioning of `cairo` in abcm2ps' Makefile is a product of autotools.

Anybody have another interpretation?

packages/abcm2ps/PKGBUILD
SpotlightKid: I've taken it out. Seems unlikely somebody still has that package installed.

packages/abcm2ps/PKGBUILD
SpotlightKid: Ok, left it in.

SpotlightKid:
Hold on, `check` fails. Probably missing fonts.

SpotlightKid:
Fixed again.

cbix (approved):
Preemptive approval, just fix the nvchecker conflicts and we're good to merge 🎉

PR #45, opened 2022-04-05
Package aether.lv2

cbix:
[**Aether**](https://github.com/Dougal-s/Aether)

## TODO
- [x] add roboto ttf dependency and link from the .lv2 dir
  - pro: smaller install size if user already has roboto fonts installed
  - con: .lv2 is more system-dependent (should be ok for system-installed plugins tho)
  - con: namcap throws an error about this
- [x] fix dependencies
- [x] install license
- [x] should we package the user manual?
- [x] fix a crash that happens when disabling IN/OUT on the spectrogram (doesn't happen when building from git)
- [x] fix build of patched version on aarch64

### DIFF
--- a/packages/aether.lv2/PKGBUILD
+++ b/packages/aether.lv2/PKGBUILD
@@ -11,7 +11,7 @@ url='https://dougal-s.github.io/Aether'
 license=(MIT)
 groups=(lv2-plugins pro-audio)
 depends=(gcc-libs libglvnd libx11)
-makedepends=(cmake glu)
+makedepends=(cmake glu md4c)
 checkdepends=(lv2lint)
 optdepends=('lv2-host: for running the plugin')
 _pugl_ref=9fd2cd2
@@ -43,6 +43,9 @@ build() {
     -DCMAKE_INSTALL_PREFIX='/usr' \
     -Wno-dev
   cmake --build build
+
+  cd $_projectname-$pkgver/usermanual
+  md2html -f --github USERMANUAL.md > index.html
 }
 
 check() {
--- a/packages/aether.lv2/PKGBUILD
+++ b/packages/aether.lv2/PKGBUILD
@@ -1,0 +1,66 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+_projectname=Aether
+pkgname=aether.lv2
+pkgver=1.2.1
+pkgrel=1
+pkgdesc='An algorithmic reverb LV2 plugin based on Cloudseed'
+arch=(x86_64 aarch64)
+url='https://dougal-s.github.io/Aether'
+license=(MIT)
+groups=(lv2-plugins pro-audio)
+depends=(gcc-libs libglvnd libx11)
+makedepends=(cmake glu)
+checkdepends=(lv2lint)
+optdepends=('lv2-host: for running the plugin')
+_pugl_ref=9fd2cd2
+_nanovg_ref=077b65e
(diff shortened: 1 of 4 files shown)


packages/aether.lv2/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+noextract=("pugl-$_pugl_ref.tar.gz" "nanovg-$_nanovg_ref.tar.gz")
+
+prepare() {
+  cd $_projectname-$pkgver
+  # extract submodules
+  bsdtar -xf ../pugl-$_pugl_ref.tar.gz -C extern/pugl --strip-components 1
+  bsdtar -xf ../nanovg-$_nanovg_ref.tar.gz -C extern/nanovg --strip-components 1
+
+  # A crash is already fixed in master
+  patch -Np1 -i ../checkout-1162d0e.patch
```
cbix, line 37: This patch is quite big but I didn't notice any breaking changes from a user perspective so I'd keep it in until the next release.

packages/aether.lv2/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+
+  cd $_projectname-$pkgver
+  install -vDm644 LICENSE.md -t "$pkgdir"/usr/share/licenses/$pkgname
+  install -vDm644 usermanual/* -t "$pkgdir"/usr/share/doc/$pkgname
+
+  cd "$srcdir"/build/$pkgname
+  install -vDm755 *.so -t "$pkgdir"/usr/lib/lv2/$pkgname
+  install -vDm644 *.ttl -t "$pkgdir"/usr/lib/lv2/$pkgname
+  install -d "$pkgdir"/usr/lib/lv2/$pkgname/fonts
+  ln -vs /usr/share/fonts/TTF/Roboto-{Light,Regular}.ttf -t "$pkgdir"/usr/lib/lv2/$pkgname/fonts
```
cbix, line 65: Is this legal? :P

packages/aether.lv2/PKGBUILD
SpotlightKid: It only seems to change things about the building process. So if it fixes a crash, it seems justified to include it.

packages/aether.lv2/PKGBUILD
SpotlightKid: As long as Robocop doesn't see it...

packages/aether.lv2/PKGBUILD
cbix: The patch is not just that commit 1162d0e but all changes since v1.2.1 up to (including) 1162d0e.

packages/aether.lv2/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+    -t "UI Symbols" \
+    http://github.com/Dougal-s/Aether
+}
+
+package() {
+  depends+=(ttf-roboto)
+
+  cd $_projectname-$pkgver
+  install -vDm644 LICENSE.md -t "$pkgdir"/usr/share/licenses/$pkgname
+  install -vDm644 usermanual/* -t "$pkgdir"/usr/share/doc/$pkgname
```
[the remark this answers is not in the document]
cbix: Good idea! Turns out `md2html` from the `md4c` package (which also has no extra dependencies) renders the lists correctly, but doesn't seem to have an option to make the TOC work, I think that's ok for now.

08c8c4a3217d4dc15757da678b14fc274db5eecf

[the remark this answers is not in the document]
SpotlightKid: Yeah, good enough ;)

SpotlightKid (approved):

PR #50, opened 2022-04-20
Package jack-matchmaker

cbix:


diovudau (approved):
The AUR pkgbuild always worked and is from @SpotlightKid  anyway. 
Should be good.

PR #51, opened 2022-04-26
package glasscoder

thisven:
[GlassCoder](https://github.com/ElvishArtisan/GlassCoder)

Refering to [INSTALL](https://github.com/ElvishArtisan/GlassCoder/blob/master/INSTALL) codec and source dependencies could be reduced if using at least one of each. And as there is no package for [AudioScience HPI driver](https://www.audioscience.com/internet/download/linux_drivers.htm) yet, the package was built without support for it.


cbix:
@thisven is the HPI driver open source?

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: this.ven <misgi@ven.uber.space>
+
+pkgname=glasscoder
+pkgver=1.0.4
+pkgrel=1
+pkgdesc="Minimalist audio encoder for IceCast and Shoutcast streaming
+servers as well as HTTP Live Streaming (HLS)"
```
SpotlightKid, line 8: No line-breaks in `pkgdesc`, Use single quotes.

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+arch=(x86_64)
+url="https://github.com/ElvishArtisan/GlassCoder"
```
SpotlightKid, line 10: Use single quotes.

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+license=(GPL2)
+depends=(alsa-lib jack2 curl lame libfdk-aac libogg libsamplerate libsndfile libvorbis openssl opus python qt5-tools taglib twolame)
+makedepends=(docbook-xsl docbook5-xml libxslt)
+source=(https://github.com/ElvishArtisan/GlassCoder/releases/download/v1.0.4/glasscoder-1.0.4.tar.gz)
```
SpotlightKid, line 14: Use:

    source("https://github.com/ElvishArtisan/GlassCoder/releases/download/v$pkgver/$pkgname-$pkgver.tar.gz")

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+sha256sums=('350714c1cf23caf1e2f12431385a946ef8ca4a585329a399cebcf5ad598ee7c8')
+
+build() {
+  export DOCBOOK_STYLESHEETS="/usr/share/xml/docbook/xsl-stylesheets-1.79.2"
```
SpotlightKid, line 18: put the this line directly before the `:/configure` line.

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+  cd "$pkgname-$pkgver"
```
SpotlightKid, line 19: No quotes necessary here.

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+
```
SpotlightKid, line 20: No empty line here.

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+  ./configure --prefix=/usr
+  make
+}
+
+package() {
+  cd "$pkgname-$pkgver"
```
SpotlightKid, line 26: No quotes necessary here.

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+
```
SpotlightKid, line 27: No empty line here.

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+  make DESTDIR="$pkgdir" install
+}
```
SpotlightKid, line 29: File should end with a newline.

SpotlightKid (changes requested):
Some purely style-based review comments. I haven't tried to build the package myself yet.

packages/glasscoder/PKGBUILD
cbix: Also, according to https://wiki.archlinux.org/title/PKGBUILD#pkgdesc
> This is recommended to be 80 characters or less

Maybe sth like `Minimalist audio encoder for IceCast, Shoutcast and HLS`

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,29 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: this.ven <misgi@ven.uber.space>
+
+pkgname=glasscoder
+pkgver=1.0.4
+pkgrel=1
+pkgdesc="Minimalist audio encoder for IceCast and Shoutcast streaming
+servers as well as HTTP Live Streaming (HLS)"
+arch=(x86_64)
```
cbix, line 9:
```suggestion
arch=(x86_64 aarch64)
```

Since fd11c14296449542c90300129c46b223881d5003 the CI doesn't build aarch64 automatically, so I'd recommend to first always add it explicitly and if the build fails, try to fix or just remove it again 🤔

packages/glasscoder/PKGBUILD
cbix: Isn't it also bad practice to `export` environment variables in a PKGBUILD? Not 100 % sure about this now but maybe we should do
```
DOCBOOK_STYLESHEETS="..." ./configure ...

**Comment by thisven** [comment]
> @thisven is the HPI driver open source?

Yes, it is:

> The driver and tools are supplied as source under the GPL license. Userspace libraries are under a BSD style license. Both 32 and 64 bit kernels are supported. (Note: mixed 32 bit user / 64 bit kernel is not supported)

and it says that:

> [ALSA source](http://alsa-project.org/main/index.php/Download) includes snd-asihpi driver and firmware, however this will lag the latest HPI drivers mentioned above.

Quoted from[ AudioScience-Linux Drivers](https://www.audioscience.com/internet/download/linux_drivers.htm) page.

File: packages/glasscoder/PKGBUILD (hunk 1 above)
Comment (thisven) [reply packages/glasscoder/PKGBUILD]: You're right. I've shortened that.

File: packages/glasscoder/PKGBUILD (hunk 4 above)
Comment (thisven) [reply packages/glasscoder/PKGBUILD:18]: Unfurtunately that doesn't work. Compilation fails because of missing variable.

File: packages/glasscoder/PKGBUILD (hunk 10 above)
Comment (thisven) [reply packages/glasscoder/PKGBUILD]: It works. :smiley:

**Comment by thisven** [comment]
Builds for aarch64 are now also included. Thanks @cbix.

File: packages/glasscoder/PKGBUILD (hunk 4 above)
Comment (SpotlightKid) [reply packages/glasscoder/PKGBUILD:18]: Probably it is also needed in the environment of `make`. `export`ing in the `PKGBUILD` is fine, IMO. It only affects the shell context of `makepkg` and sub-processes.

Comment (cbix) [reply packages/glasscoder/PKGBUILD:18]: Ok, I just remember reading that somewhere but I agree with you :D

**Comment by thisven** [comment]
[Commits](/osam-cologne/archlinux-proaudio/commit/bf26f093e8ed2895466af3f3e57eaef433b37f61) for nvchecker are now also done.

**Comment by cbix** [comment]
Also take a quick look at the [namcap output](https://ci.cbix.de/osam-cologne/archlinux-proaudio/422/1/6) (you can also run this locally on the built package):

The first part is not critical, the CI doesn't have any packages installed at this point, so we can ignore it:
```
glasscoder W: Referenced library 'python' is an uninstalled dependency
glasscoder W: Referenced library 'libtag.so.1' is an uninstalled dependency
[...]
```
Then an error:
```
glasscoder E: Dependency hicolor-icon-theme detected and not included (needed for hicolor theme hierarchy)
```
The [`hicolor-icon-theme`](https://archlinux.org/packages/extra/any/hicolor-icon-theme/) package needs to be a package dependency because it includes the predefined directory structure for included icons.

Then more warnings:
```
glasscoder W: Dependency included and not needed ('alsa-lib')
glasscoder W: Dependency included and not needed ('jack2')
glasscoder W: Dependency included and not needed ('curl')
[...]
```
This may or may not mean anything, but it's always a good idea to reflect on what is strictly required at runtime and what's just a `makedepend` (needed at build time). I can imagine curl being required but `jack` and `alsa-lib` being optional.

File: packages/glasscoder/PKGBUILD [hunk 11]
```diff
@@ -0,0 +1,26 @@
...
+
+pkgname=glasscoder
+pkgver=1.0.4
+pkgrel=1
+pkgdesc='Minimalist audio encoder for IceCast, Shoutcast and HLS'
+arch=(x86_64 aarch64)
+url='https://github.com/ElvishArtisan/GlassCoder'
+license=(GPL2)
+depends=(alsa-lib jack2 curl lame libfdk-aac libogg libsamplerate libsndfile libvorbis openssl opus python qt5-tools taglib twolame)
+makedepends=(docbook-xsl docbook5-xml libxslt)
```
cbix, line 12: There you go, an example of optional dependencies:
```suggestion
depends=(curl lame libfdk-aac libogg libsamplerate libsndfile libvorbis openssl opus python qt5-tools taglib twolame)
makedepends=(alsa-lib docbook-xsl docbook5-xml jack libxslt)
optdepends=('alsa-lib: for ALSA audio sources'
            'jack: for JACK audio sources')
```

cbix: Oh, I might be wrong here, the software doesn't seem to start at all if some dependencies are not installed ...

thisven: I mostly agree with that, but what's about the [hicolor-icon-theme](https://archlinux.org/packages/extra/any/hicolor-icon-theme/) packages you mentioned above?

thisven: Shouldn't we include that, too?

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+pkgver=1.0.4
+pkgrel=1
+pkgdesc='Minimalist audio encoder for IceCast, Shoutcast and HLS'
+arch=(x86_64 aarch64)
+url='https://github.com/ElvishArtisan/GlassCoder'
+license=(GPL2)
+depends=(curl lame libfdk-aac libogg libsamplerate libsndfile libvorbis openssl opus python qt5-tools taglib twolame)
+makedepends=(alsa-lib docbook-xsl docbook5-xml jack libxslt)
+optdepends=('alsa-lib: for ALSA audio sources'
+            'jack: for JACK audio sources')
```
cbix, line 14: @thisven Sorry, I just checked and my earlier suggestion was wrong. `jack` and `alsa-lib` are strictly required at runtime, otherwise you get errors like
```
glasscoder: error while loading shared libraries: libjack.so.0: cannot open shared object file: No such file or directory
```
And of course I forgot the `hicolor-icon-theme`. There you go:
```suggestion
depends=(alsa-lib curl hicolor-icon-theme jack lame libfdk-aac libogg libsamplerate libsndfile
  libvorbis openssl opus python qt5-tools taglib twolame)
makedepends=(docbook-xsl docbook5-xml libxslt)
```

cbix: (The same would be the case if we built with libhpi/AudioScience HPI support 😅)

thisven: OK, then almost back to the first approach, but including the icon theme package. :sweat_smile:

cbix (approved):

SpotlightKid (changes requested):
Here are changes for better dependency declararions:

``` diff
diff --git a/packages/glasscoder/PKGBUILD b/packages/glasscoder/PKGBUILD
index b4dd558..c0c6e0f 100644
--- a/packages/glasscoder/PKGBUILD
+++ b/packages/glasscoder/PKGBUILD
@@ -8,8 +8,16 @@ pkgdesc='Minimalist audio encoder for IceCast, Shoutcast and HLS'
 arch=(x86_64 aarch64)
 url='https://github.com/ElvishArtisan/GlassCoder'
 license=(GPL2)
-depends=(alsa-lib curl hicolor-icon-theme jack lame libfdk-aac libogg libsamplerate libsndfile libvorbis openssl opus python qt5-tools taglib twolame)
-makedepends=(docbook-xsl docbook5-xml libxslt)
+depends=(curl hicolor-icon-theme openssl qt5-tools taglib)
+makedepends=(alsa-lib flac jack docbook-xsl docbook5-xml lame libfdk-aac
+             libogg libsamplerate libsndfile libvorbis libxslt opus python
+             zstd twolame)
+optdepends=('lame: for MP3 output support'
+            'libfdk-aac: for AAC output support'
+            'python: for rivendell integration'
+            'python-requests: for rivendell integration'
+            'rivendell: for rivendell integration'
+            'twolame: for MP2 output support')
 source=(https://github.com/ElvishArtisan/GlassCoder/releases/download/v$pkgver/$pkgname-$pkgver.tar.gz)
 sha256sums=('350714c1cf23caf1e2f12431385a946ef8ca4a585329a399cebcf5ad598ee7c8')
 
@@ -21,6 +29,7 @@ build() {
 }
 
 package() {
+  depends+=(libasound.so libFLAC.so libjack.so libogg.so libopus.so libsamplerate.so libsndfile.so libvorbis.so libvorbisenc.so libzstd.so)
   cd $pkgname-$pkgver
   make DESTDIR="$pkgdir" install
 }
``` 

Explanation: MP3, MP2 and AAC is loaded dynamically via `dlopen()` at run-time, so the libraries providing support for these can be optional dependencies.

Rivendell integration: this is provided by a python script,w hich import `requests` and a python module from the `rivdendell` package, so `python`, `python-requests` and `rivendell` can be optional dependencies too. `rivendell` curently exists only in the AUR, though, so we should either package it as well or not install the rivdendell integration (see ? configure` options).

The rest of the changes concern depending on the soname instead of the package that provides it.

cbix:
If rivendell still requires qt4, we can't easily package it...

SpotlightKid:
The `configure` script errors out, if Python is not found, even though it says it is optional. I can't find a way to disable the check without patching `configure.ac`

cbix:
@SpotlightKid let's keep it as a `makedepend` then :)

- not sure why you included `zstd`/`libzstd`, it's not being used by glasscoder itself and is also present on every Arch system by transient dependency base → pacman → libarchive → zstd
- `libopus`, `libogg`, `libvorbis` and `libvorbisenc` [are also `dlopen`ed](https://github.com/ElvishArtisan/GlassCoder/search?q=dlopen) and thus optional, plus transitively required by `libsndfile`, so we can safely leave them out (make optional)
- `libFLAC` also seems to be linked only through `libsndfile`, so it's not even an optional dependency IMO
- for future updates let's keep track of https://github.com/ElvishArtisan/GlassCoder/issues/15
- regarding `rivendell`: we could leave the python things in the package but comment out the `optdepends` until we eventually package rivendell. That way the package itself won't change later, just the `optdepends`. What do you think?

result:
```diff
diff --git a/packages/glasscoder/PKGBUILD b/packages/glasscoder/PKGBUILD
index b4dd558..c0c6e0f 100644
--- a/packages/glasscoder/PKGBUILD
+++ b/packages/glasscoder/PKGBUILD
@@ -8,8 +8,16 @@ pkgdesc='Minimalist audio encoder for IceCast, Shoutcast and HLS'
 arch=(x86_64 aarch64)
 url='https://github.com/ElvishArtisan/GlassCoder'
 license=(GPL2)
-depends=(alsa-lib curl hicolor-icon-theme jack lame libfdk-aac libogg libsamplerate libsndfile libvorbis openssl opus python qt5-tools taglib twolame)
-makedepends=(docbook-xsl docbook5-xml libxslt)
+depends=(curl hicolor-icon-theme openssl qt5-tools taglib)
+makedepends=(alsa-lib flac jack docbook-xsl docbook5-xml lame libfdk-aac
+             libogg libsamplerate libsndfile libvorbis libxslt opus python
+             twolame)
+optdepends=('lame: for MP3 output support'
+            'libfdk-aac: for AAC output support'
+            'python: for rivendell integration'
+            'python-requests: for rivendell integration'
+            'rivendell: for rivendell integration'
+            'twolame: for MP2 output support')
 source=(https://github.com/ElvishArtisan/GlassCoder/releases/download/v$pkgver/$pkgname-$pkgver.tar.gz)
 sha256sums=('350714c1cf23caf1e2f12431385a946ef8ca4a585329a399cebcf5ad598ee7c8')
 
@@ -21,6 +29,7 @@ build() {
 }
 
 package() {
+  depends+=(libasound.so libjack.so libsamplerate.so libsndfile.so)
   cd $pkgname-$pkgver
   make DESTDIR="$pkgdir" install
 }
```

ElvishArtisan:
> And as there is no package for [AudioScience HPI driver](https://www.audioscience.com/internet/download/linux_drivers.htm) yet, the package was built without support for it.

FWIW, there is Debian packaging information being maintained for this (currently tracking Ubuntu 20.04 Focal Fosa) at https://github.com/ElvishArtisan/debian-hpklinux.

thisven:
@cbix I like your suggestion and thanks for contacting @ElvishArtisan

thisven:
> FWIW, there is Debian packaging information being maintained for this (currently tracking Ubuntu 20.04 Focal Fosa) at https://github.com/ElvishArtisan/debian-hpklinux.

Thanks for your hint. Maybe we can use that for #52.

packages/glasscoder/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+sha256sums=('350714c1cf23caf1e2f12431385a946ef8ca4a585329a399cebcf5ad598ee7c8')
+
+build() {
+  cd $pkgname-$pkgver
+  export DOCBOOK_STYLESHEETS="/usr/share/xml/docbook/xsl-stylesheets-1.79.2"
+  ./configure --prefix=/usr
+  make
+}
+
+package() {
```
cbix, line 34:
```suggestion
package() {
  depends+=(libasound.so libjack.so libsamplerate.so libsndfile.so)
```

thisven:
As PR #52 is closed, how about including the hpklinux package as dependency?

```
diff --git a/packages/glasscoder/PKGBUILD b/packages/glasscoder/PKGBUILD
index b9aa575..280d4ba 100644
--- a/packages/glasscoder/PKGBUILD
+++ b/packages/glasscoder/PKGBUILD
@@ -9,10 +9,11 @@ arch=(x86_64 aarch64)
 url='https://github.com/ElvishArtisan/GlassCoder'
 license=(GPL2)
 depends=(curl hicolor-icon-theme openssl qt5-tools taglib)
-makedepends=(alsa-lib flac jack docbook-xsl docbook5-xml lame libfdk-aac
+makedepends=(alsa-lib flac hpklinux jack docbook-xsl docbook5-xml lame libfdk-aac
              libogg libsamplerate libsndfile libvorbis libxslt opus python
              twolame)
-optdepends=('lame: for MP3 output support'
+optdepends=('hpklinux: for AudioScience HPI driver support'
+            'lame: for MP3 output support'
             'libfdk-aac: for AAC output support'
             'libogg: for OGG Opus/Vorbis output support'
             'libvorbis: for OGG Vorbis output support'
```

thisven:
By this we got all features and supported devices covered, but on the other hand I don't know if it's according to [KISS](https://wiki.archlinux.org/title/Arch_terminology#KISS). To me it means that we should avoid pulling in packages.

cbix:
I personally don't see strong reasons not to include hpklinux. It's only around 1 MB extra and maybe eventually can be made optional anyway.

thisven:
I agree. But how can we make it optional?

cbix:
> how can we make it optional?

We can't right now, but maybe eventually https://github.com/ElvishArtisan/GlassCoder/issues/15 gets addressed ;)

thisven:
That would be sweet. :heart_eyes:

cbix:
Sure, but I agree with the author that the benefits are more idealistic than practical, and it would increase code complexity on their side. But I leave the decision to you whether we package with ASIHPI support or not :)

thisven:
I'd merge this branch into master at the end of the week, if there are no contradictions.

SpotlightKid (approved):

PR #52, opened 2022-04-27
package hpklinux

cbix:
[AudioScience HPI user-space libs](https://www.audioscience.com/internet/download/linux_drivers.htm)

Can't practically test any of this because I don't own their hardware.

**TODO**
- [x] fix namcap warnings/errors
- [x] do we need to build alsa-drv or is that included in the kernel already?
  - yes, the snd-asihpi module is included in the kernel
- [x] fix aarch64 build
  - Archlinux ARM kernel does not include drivers and it's unlikely that there are any users for this on ARM, but hey
  - fails because some structs are padded based on `HPI_BUILD_64BIT` flag which is only enabled for x86_64 so a compile-time size assertion fails
- [x] correct license??
  - uses a Zlib/BSD style license, which is GPLv2-compatible, but there is no explicit `LICENSE` file, we may just ship one ourselves...
- [x] figure out nvchecker rule for this


thisven:
What about converting the [Ubuntu package](https://launchpad.net/~audioscience/+archive/ubuntu/release) with a tool like [debtap](https://github.com/helixarch/debtap)?

cbix:
> What about converting the [Ubuntu package](https://launchpad.net/~audioscience/+archive/ubuntu/release) with a tool like [debtap](https://github.com/helixarch/debtap)?

If I understand correctly, debtap converts a package and does  not create a PKGBUILD. Anyway, my PKGBUILD for hpklinux is basically finished. The only thing left is a license file, which the original source package does not provide (license text is part of the code).

thisven:
I see. What about submitting a ticket about that? Or are allowed to package a copy of GPL and BSD license as those are mentioned at the official support page?

diovudau:
> > What about converting the [Ubuntu package](https://launchpad.net/~audioscience/+archive/ubuntu/release) with a tool like [debtap](https://github.com/helixarch/debtap)?
> 
> If I understand correctly, debtap converts a package and does not create a PKGBUILD. Anyway, my PKGBUILD for hpklinux is basically finished. The only thing left is a license file, which the original source package does not provide (license text is part of the code).

If it is a standard license we can just use the arch standard licenses.  A LICENSE file is not needed for a software project anyway, it is just convenience.  Licenses apply to code files.   The current case is less problematic than software packages that do not have license headers in the code but only a single LICENSE file in the root.

cbix:
> If it is a standard license we can just use the arch standard licenses

Not really, some parts are GPLv2 and some use customized Zlib license in their header:

> Copyright (C) 1997-2017 AudioScience, Inc. All rights reserved.
> 
> This software is provided 'as-is', without any express or implied warranty.
> In no event will AudioScience Inc. be held liable for any damages arising
> from the use of this software.
> 
> Permission is granted to anyone to use this software for any purpose,
> including commercial applications, and to alter it and redistribute it
> freely, subject to the following restrictions:
> 
> 1. The origin of this software must not be misrepresented; you must not
>    claim that you wrote the original software. If you use this software
>    in a product, an acknowledgment in the product documentation would be
>    appreciated but is not required.
> 2. Altered source versions must be plainly marked as such, and must not be
>    misrepresented as being the original software.
> 3. This copyright notice and list of conditions may not be altered or removed
>    from any source distribution.
> 
> AudioScience, Inc. <support@audioscience.com>
> 
> ( This license is GPL compatible see http://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses )

I added a way to extract the license during packaging now in 5ddd44161437933dafdebf7f495fd04ab1051a93, guess that will do it :)

diovudau (approved):
Yes, licensing is clear enough now.

PR #56, opened 2022-04-30
fix: soname dependencies; formatting conventions

SpotlightKid:


cbix:
also noticed `groups=(pro-audio)` is missing for `klick` ;)

PR #62, opened 2022-05-16
package rtcqs

cbix:
https://codeberg.org/rtcqs/rtcqs

I may create another package for the `python-pysimplegui` optdepend, however that's not really proaudio-related ;)

### DIFF
--- a/packages/rtcqs/PKGBUILD
+++ b/packages/rtcqs/PKGBUILD
@@ -1,0 +1,28 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+pkgname=rtcqs
+pkgdesc='Scanner for potential real-time bottlenecks'
+pkgver=0.5.3
+pkgrel=1
+arch=(any)
+url='https://codeberg.org/rtcqs/rtcqs'
+license=(MIT)
+depends=(python)
+makedepends=(python-build python-installer python-setuptools python-wheel)
+optdepends=('python-pysimplegui: for rtcqs_gui')
+groups=(pro-audio)
+source=("https://files.pythonhosted.org/packages/source/${pkgname::1}/$pkgname/$pkgname-$pkgver.tar.gz")
+sha256sums=('34baf6eeb87660c8cd975221001e10c8aca73e2bf0474ca8de1afd63555f3fdd')
+
+build() {
+  cd $pkgname-$pkgver
+  python -m build --wheel --no-isolation
+}
+
+package() {
+  cd $pkgname-$pkgver
+  python -m installer --destdir="$pkgdir" dist/*.whl
+  # license
+  install -Dm644 LICENSE -t "$pkgdir"/usr/share/licenses/${pkgname}
+}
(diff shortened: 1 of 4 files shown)


SpotlightKid:
> I may create another package for the `python-pysimplegui` optdepend

I assume you have seen https://aur.archlinux.org/packages/python-pysimplegui

packages/rtcqs/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+build() {
+  cd $pkgname-$pkgver
+  python -m build --wheel --no-isolation
+}
+
+package() {
+  cd $pkgname-$pkgver
+  python -m installer --destdir="$pkgdir" dist/*.whl
+  # license
+  install -Dm644 LICENSE -t "$pkgdir"/usr/share/licenses/${pkgname}
```
SpotlightKid, line 27: Tiny nitpick: `$pkgname`

SpotlightKid (changes requested):
The package installs the file

    /usr/lib/python3.10/site-packages/rtcqs/rtcqs_qt_gui.py

But when I try to run it, I get:

```
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.10/site-packages/rtcqs/rtcqs_qt_gui.py", line 3, in <module>
    import PySimpleGUIQt as sg
ModuleNotFoundError: No module named 'PySimpleGUIQt'
```

Maybe it's worth an upstream ticket.

When I run `rtcqs_gui` it shows the following warning on the "Audio Group" tab:

> User chris is currently not in the audio group. Add yourself to the audio group with 'sudo usermod -a -G audio chris' and log in again. See also https://wiki.linuxaudio.org/wiki/system_configuration#audio_group

The correct group name to check on Arch Linux would be `realtime`.

Maybe also worth an upstream ticket.

SpotlightKid (commented):
Does upstream plan a new release including the `.desktop` file soon? Otherwise it might be worth adding the file from the repo to the `source` array and installing it manually.

packages/rtcqs/PKGBUILD
cbix: f5f1a65e3ddce451167c8f7d7ed70dd56a7f6b94

cbix:
The Qt GUI is considered beta and neither included as a console script under `/usr/bin` nor installed as executable. I'd say we should not expect users trying to run that. If you want I can just remove it during build.

cbix:
> The correct group name to check on Arch Linux would be `realtime`.

Interesting, do you have a reference for this? For some reason I have both groups on my system.

SpotlightKid:
* https://wiki.archlinux.org/title/Realtime_process_management
* https://archlinux.org/packages/community/any/realtime-privileges/

SpotlightKid:
> If you want I can just remove it during build.

I think it's ok to just leave it there. It's a 0.x release after all.

SpotlightKid (approved):
LGTM

cbix:
@SpotlightKid I opened this issue now https://codeberg.org/rtcqs/rtcqs/issues/4

PR #65, opened 2022-05-22
Automatically push selected packages to AUR

cbix:
We often add packages in this repo that don't exist on the AUR yet, so idea is to use the AUR as a "mirror" making it easier for users to discover our packages. If a package is listed in `aur/packages`, all changes to the package that are pushed to `master` will also be committed to the AUR.

**TODO**
- [x] fix git permissions
- [x] add `aur/README.md`
- [x] limit drone step to `master`


tools/push-aur.sh
```diff
@@ -0,0 +1,35 @@
...
+    cd $PKG
+
+    # generate .SRCINFO
+    sudo -Eu nobody makepkg --printsrcinfo > .SRCINFO
+
+    # update .gitignore
+    cat "$ROOT"/aur/gitignore-base >> .gitignore
+
+    # commit and push
+    git add .
```
SpotlightKid, line 31: I think we have to be careful here. In this repo some files in `/packages/*/` are excluded by the top-level `.gitignore`, but the AUR package repo might not have a proper `.gitignore` to exclude them as well.

cbix: That's why I'm appending the contents of `gitignore-base` above.

Anyway, what is the common practice for `.gitignore`s in the AUR? I often see maintainers using negated ignores to selectively include files and ignore everything else. Maybe we could generate the `.gitignore` from scratch here using `git ls-files packages/$PKG/` and a static list of common ignores.

SpotlightKid: > That's why I'm appending the contents of gitignore-base above.

Ah, I overlooked that.

> Anyway, what is the common practice for .gitignores in the AUR?

I don't know. I usually just exclude:

    src/
    pkg/
    *.pkg.tar.xz
    *.pkg.tar.zst
    <package>.*.tar.gz

SpotlightKid:
I don't have time to look at this right now. I'll try tomorrow evening.

SpotlightKid (commented):
LGTM. I guess we have to see how this works out in practice.

SpotlightKid (approved):

PR #66, opened 2022-05-22
feat: `surge-xt` package

Bleuzen:
This packages the Surge XT synth:
https://surge-synthesizer.github.io/

PKGBUILD is based on `surge` package from official repos

NOTE: this does currently not build the LV2

### DIFF
--- a/packages/surge-xt/PKGBUILD
+++ b/packages/surge-xt/PKGBUILD
@@ -1,0 +1,103 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Bleuzen <supgesu at gmail dot com>
+# Original author of surge package: David Runge <dvzrv@archlinux.org>
+
+# This currently only builds VST3 plugins
+# lv2 would need custom JUCE:
+# https://github.com/lv2-porting-project/JUCE
+# or simply wait for JUCE 7 release which will officially support lv2
+
+_name="Surge XT"
+_repo_name="surge"
+pkgname=surge-xt
+pkgver=1.0.1
+pkgrel=2
+pkgdesc="An Open Source Digital Synthesizer"
+arch=('x86_64')
+url="https://surge-synthesizer.github.io/"
+license=('GPL3')
+groups=('pro-audio' 'vst3-plugins')
+# groups=('lv2-plugins' 'pro-audio' 'vst3-plugins')
+depends=('cairo' 'fontconfig' 'gcc-libs' 'glibc' 'libx11' 'libxcb'
+'libxkbcommon-x11' 'xcb-util' 'xcb-util-cursor' 'xcb-util-image'
+'xcb-util-keysyms' 'xcb-util-renderutil' 'xdg-utils')
+makedepends=('cmake' 'freetype2' 'libsndfile' 'python')
+# makedepends=('cmake' 'freetype2' 'libsndfile' 'lv2' 'python')
+optdepends=(
+  # 'lv2-host: for the LV2 plugin'
+  'vst3-host: for the VST3 plugin'
+)
+# checkdepends=('lv2lint')
+source=("${pkgname}-${pkgver}.tgz::https://github.com/surge-synthesizer/releases-xt/releases/download/${pkgver}/surge-src-${pkgver}.tar.gz")
+sha256sums=('100b536f741c62529800fc26e3773567591385d4152db305382b4315f6e39c55')
+
+prepare() {
+  mv -v "$_repo_name" "$pkgname-$pkgver"
+  cd "$pkgname-$pkgver"
+  # first shot at improving reproducibility:
+  # https://github.com/surge-synthesizer/surge/issues/2389
+  sed -e 's/@SURGE_BUILD_DATE@/0/g' \
+      -e 's/@SURGE_BUILD_TIME@/0/g' \
+      -e 's/@SURGE_BUILD_HASH@/0/g' \
+      -e 's/@SURGE_BUILD_FQDN@/None/g' \
+      -e "s/@SURGE_BUILD_ARCH@/${CARCH}/g" \
+      -e "s/@SURGE_FULL_VERSION@/${pkgver}/g" \
+      -i src/common/version.cpp.in
+}
+
+build() {
+  cd "$pkgname-$pkgver"
+  # binaries and shared libs have executable stack:
+  # https://github.com/surge-synthesizer/surge/issues/1701
+  export LDFLAGS="${LDFLAGS},-z,noexecstack"
+  # build surge-headless
+  cmake -DCMAKE_INSTALL_PREFIX='/usr' \
+        -DCMAKE_BUILD_TYPE='None' \
+        -W no-dev \
+        -B build \
+        -S .
+  make VERBOSE=1 -C build
+}
+
+# check() {
+#   cd "$pkgname-$pkgver"
+#   lv2lint -Mpack -I "build-test/surge_xt_products/${_name}.lv2" \
+#     "https://surge-synthesizer.github.io/lv2/surge" \
+#     || echo "Known to fail: https://github.com/surge-synthesizer/surge/issues/2392"
+# 
+#   mkdir -vp test_data/${pkgname}
+#   cp -av resources/data/* test_data/${pkgname}
+#   XDG_DATA_HOME=test_data build/surge-headless
+# }
+
+package() {
+  depends+=('libfreetype.so')
+  cd "$pkgname-$pkgver"
+
+  # install lv2
+#   install -vDm 755 "build/surge_xt_products/${_name}.lv2/${_name}.so" \
+#     -t "${pkgdir}/usr/lib/lv2/${_name}.lv2/"
+#   install -vDm 644 "build/surge_xt_products/${_name}.lv2/"*.ttl \
+#     -t "${pkgdir}/usr/lib/lv2/${_name}.lv2/"
+#   install -vDm 755 "build/surge_xt_products/${_name} Effects.lv2/${_name} Effects.so" \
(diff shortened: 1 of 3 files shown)


cbix:
Nice, thanks for this!

I'll try to take some time later and comment a bit on PKGBUILD code style and improvements. Just noticed the build is failing because it's missing a header `X11/extensions/Xrandr.h`. Let's find the missing dependency:
```
$ pacman -Qo /usr/include/X11/extensions/Xrandr.h
/usr/include/X11/extensions/Xrandr.h is owned by libxrandr 1.5.2-3
```

Bleuzen:
> Just noticed the build is failing because it's missing a header `X11/extensions/Xrandr.h`. Let's find the missing dependency

I added some more dependencies that it complained about to `makedepends`. Not sure if some of them need to go to `depends` tho, can't really test rn since I don't have such a clean system.
Also can't make sure that build works now since when building this in chroot it completely fills up my 32 gb of ram and freezes my machine

cbix:
> sorry, maybe I shouldn't have pushed because of that, hope it doesn't kill your CI server

Don't worry, it's all running in docker containers with generous timeouts. Not sure why it would fill up your RAM, however the way we're building is a bit different to archroot, instead we're using the `archlinux/archlinux:base-devel` docker image for a clean build environment.
You could even add `aarch64` to the `arch()` array to test in this PR if the CI can build this. If that fails, simply remove aarch64 again.

> can't really test rn since I don't have such a clean system

Yea, one thing I'll eventually set up is a temporary storage for WIP built packages so one can simply download them after CI finishes building from a PR...

1. I noticed the binaries contain a space (for example `usr/bin/Surge XT`), is there a way to rename them to for example `usr/bin/surge-xt`? Or will that break any functionality?
2. [Namcap](https://wiki.archlinux.org/title/Namcap) complains about some wrong build flags:
    ```
    surge-xt W: ELF file ('usr/bin/Surge XT') lacks FULL RELRO, check LDFLAGS.
    surge-xt W: ELF file ('usr/bin/Surge XT Effects') lacks FULL RELRO, check LDFLAGS.
    surge-xt W: ELF file ('usr/bin/Surge XT') lacks PIE.
    surge-xt W: ELF file ('usr/bin/Surge XT Effects') lacks PIE.
    ```
    it usually means that the linking step for these binaries is not using the LDFLAGS provided from outside (/etc/makepkg.conf), or the build script is overwriting these. If you can't fix it, just ignore that warning^^
3. would be cool to provide a simple `.desktop` file with the package (for example add `gendesk` as a makedepend and call that with some meaningful flags in the `prepare()` step)

Bleuzen:
> You could even add `aarch64` to the `arch()` array to test in this PR if the CI can build this. If that fails, simply remove aarch64 again.

Done now, will see how this goes.


>     1. I noticed the binaries contain a space (for example `usr/bin/Surge XT`), is there a way to rename them to for example `usr/bin/surge-xt`? Or will that break any functionality?

Of course these could be renamed. But I think doing it in this PKGBUILD is the wrong place. Imagine someone using these exact binary names in a script, comming from another distribution and wondering why it is broken on Arch. Doesn't really make sense. Consistency should be valued. Official packages provided by upstream use these exact names, so for better compatibility I recommend leaving the names as is. If you don't like them, open an issue on upstream to possibly get them renamed there, so all distributions could benefit from this change.

>     2. [Namcap](https://wiki.archlinux.org/title/Namcap) complains about some wrong build flags:
>        ```
>        surge-xt W: ELF file ('usr/bin/Surge XT') lacks FULL RELRO, check LDFLAGS.
>        surge-xt W: ELF file ('usr/bin/Surge XT Effects') lacks FULL RELRO, check LDFLAGS.
>        surge-xt W: ELF file ('usr/bin/Surge XT') lacks PIE.
>        surge-xt W: ELF file ('usr/bin/Surge XT Effects') lacks PIE.
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        it usually means that the linking step for these binaries is not using the LDFLAGS provided from outside (/etc/makepkg.conf), or the build script is overwriting these. If you can't fix it, just ignore that warning^^

Currently I'm not sure what to do about this

>     3. would be cool to provide a simple `.desktop` file with the package (for example add `gendesk` as a makedepend and call that with some meaningful flags in the `prepare()` step)

Same as with 1, should be added upstream so everyone could benefit and we get consistant packages for all distributions.

>     4. another warning after packaging:
>        ```
>        : Can't translate pathname 'usr/share/surge-xt/patches_3rdparty/Kinsey Dulcet/Pads/Sunrise Breakfast At The Digital Café.fxp' to UTF-8: Can't translate pathname 'usr/share/surge-xt/patches_3rdparty/Kinsey Dulcet/Pads/Wavetextures™ Pro Kit.fxp' to UTF-8: Can't translate pathname 'usr/share/surge-xt/patches_factory/Leads/µcomputer.fxp' to UTF-8
>        ```

Hm these extract to my ext4 just fine from official packages. Not sure why it doesn't like them here. Possible makepkg bug/limitation?

packages/surge-xt/PKGBUILD
```diff
@@ -0,0 +1,104 @@
...
+  # install standalone
+  install -vDm 755 "build/surge_xt_products/${_name}" \
+    -t "${pkgdir}/usr/bin/"
+  install -vDm 755 "build/surge_xt_products/${_name} Effects" \
+    -t "${pkgdir}/usr/bin/"
+
+  # install resources
+  install -vdm 755 "${pkgdir}/usr/share/${pkgname}"
+  cp -av resources/data/* "${pkgdir}/usr/share/${pkgname}"
+  install -vDm 644 {AUTHORS,README.md} -t "${pkgdir}/usr/share/doc/${pkgname}"
```
cbix, line 103: What do you think about removing the `README.md`? IMO it doesn't contain relevant information for users, but maybe we should still keep it?
```suggestion
  install -vDm 644 AUTHORS -t "${pkgdir}/usr/share/doc/${pkgname}"
```

cbix (changes requested):
Alright, I finally managed to go over your patch and test everything 🎉 Build works well and I'll see myself playing with surge-xt a bit more in the future, though it's not immediately clear to me why they renamed the project in the first place :D 

Also not sure if @dvzrv was eventually gonna migrate the existing surge package to surge-xt anyway, but either way this can be a good starting point IMO.

packages/surge-xt/PKGBUILD
```diff
@@ -0,0 +1,104 @@
...
+  'jack: backend for standalone'
+)
+# checkdepends=('lv2lint')
+source=("${pkgname}-${pkgver}.tgz::https://github.com/surge-synthesizer/releases-xt/releases/download/${pkgver}/surge-src-${pkgver}.tar.gz")
+sha256sums=('100b536f741c62529800fc26e3773567591385d4152db305382b4315f6e39c55')
+
+prepare() {
+  mv -v "$_repo_name" "$pkgname-$pkgver"
+  cd "$pkgname-$pkgver"
+  # first shot at improving reproducibility:
```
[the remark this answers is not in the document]
cbix: ok it seems to still be required

packages/surge-xt/PKGBUILD
Bleuzen: seems not useful for users of a binary package, so let's remove it

Bleuzen:
> I finally managed to go over your patch and test everything

Thanks for your suggestions :)

> though it's not immediately clear to me why they renamed the project in the first place :D

from projects readme:

> [...] the Surge Synth team undertook an effort to rebuild Surge in JUCE and add a variety of features. This project resulted in the new version of Surge, called 'Surge XT'.

The rename is because Surge and Surge XT are not interchangeable or compatible with each other. If you used Surge in older projects, it won't get replaced by Surge XT. It can't because Surge XT is a whole new plugin with new id and save format.

It is a little confusing tho since upstream put them both in one repo, maybe it would be more clear if they created a new one for Surge XT but here we are.



> Also not sure if dvzrv was eventually gonna migrate the existing surge package to surge-xt anyway

He explained over on reddit that Surge XT is currently not his priority.
(However my goal is that at some point we get this into official repos if possible.)

Also, since they are different plugins, Surge cannot be migrated into Surge XT. Surge XT needs its own package and they need to be maintained in parallel (so that you can still install Surge to be able to open older projects using this plugin).

diovudau (commented):
Why not package the README? It is part of the documentation.

cbix (approved):
Thanks! Another idea would be to add a split package for https://github.com/surge-synthesizer/surge-extra-content, to allow installing some extra skins. We could also eventually make the stock patches a split package because they take up most of the space of this package...

But for now I'm ok with merging this :)

Bleuzen:
> Why not package the README? It is part of the documentation.

Looking at the readme, it does not contain much useful stuff for our case.
The whole readme is more targeted at people seeing their git repo. It does not contain documentation about using the synth but mostly build instructions and other development stuff.
What would be the benefit of giving this information to users who downloaded an already built binary package?

Bleuzen:
> It is not up to the packager to decide what is useful and what not.

I mean, we could even ship the whole source repo with commit history and a copy of all GitHub issues in the package, because who knows xD

Seriously, you see there is a line and at some point the packager has to decide what should be in the package and what not.


> We ship as provided upstream.

There is no one and only "as provided upstream".

For example if you look at the GitHub repo, yeah, there is the readme.

But if you look in their official binary packages... they don't include this repo.

So what now? :D

I think it makes more sense to compary binary package to binary package
so it is fine here to exclude this readme (which is clearly more targeted to developers anyway)
this is also consistent with the official packages provided by upstream.

SpotlightKid:
> Of course these could be renamed. But I think doing it in this PKGBUILD is the wrong place

We could just install a symlink from `Surge XT` to `/usr/bin/surge-xt`. I find it convenient to be able to run commands from the command line without having to use key combination (shift or otherwise). But it's a minor issue.

PR #82, opened 2022-07-01
update faustlive to 2.5.11

cbix:
- [ ] fix crash


cbix:
Will continue investigating some crashes I'm experiencing after merging.

PR #83, opened 2022-07-02
package ninjamsrv

cbix:
[Ninjam](https://cockos.com/ninjam/) server
- using [my mirror](https://github.com/cbix/ninjam) because the official repo doesn't have release tags


packages/ninjamsrv/PKGBUILD
```diff
@@ -0,0 +1,44 @@
...
+pkgname=ninjamsrv
+pkgver=0.080
+pkgrel=1
+pkgdesc='Ninjam online jamming server'
+arch=(x86_64 aarch64)
+url='https://cockos.com/ninjam/'
+license=(GPL2)
+depends=(gcc-libs)
+makedepends=()
+optdepends=()
```
SpotlightKid, line 15: Can be removed?

Or should `systemd` be added? (I have no clue what other packages providing systemd service files do it.)

packages/ninjamsrv/PKGBUILD
```diff
@@ -0,0 +1,44 @@
...
+groups=(pro-audio)
+conflicts=(ninjam-server-git)
```
SpotlightKid, line 17: The correct thing would be for the `ninjam-server-git` package to have `provides=(ninjam-server)` and this one too. We can't change `ninjam-server-git`, but maybe we should add  `provides=(ninjam-server)`.

packages/ninjamsrv/PKGBUILD
```diff
@@ -0,0 +1,44 @@
...
+source=("$pkgname-$pkgver.tar.gz::https://github.com/cbix/$_name/archive/refs/tags/$pkgname/v$pkgver.tar.gz"
```
SpotlightKid, line 18: How do you determine which commits you tag as releases?

cbix: See my [comment here](https://github.com/justinfrankel/ninjam/issues/14#issuecomment-1166307136) :)

packages/ninjamsrv/PKGBUILD
cbix: Only using `provides` will lead to a conflict during installation, while explicit `conflicts` warns the user upon installation and asks about removing the conflicting package. See [the wiki](https://wiki.archlinux.org/title/PKGBUILD#provides):
> Packages providing the same item can be installed side-by-side, unless at least one of them uses a `conflicts` array.

I'll still ask the maintainer to add `provides=(ninjamsrv)`. (I used `ninjamsrv` everywhere instead of `ninjam-server` because that's how the binary is called by default and didn't want to confuse users)

packages/ninjamsrv/PKGBUILD
cbix: Some packages do (like [php7-fpm](https://archlinux.org/packages/extra/x86_64/php7-fpm/)) and others (like [gpsd](https://archlinux.org/packages/community/x86_64/gpsd/)) don't 🤔

I just removed the empty fields on this branch.

packages/ninjamsrv/PKGBUILD
SpotlightKid: Ok, thanks.

packages/ninjamsrv/PKGBUILD
SpotlightKid: Thinking some more about this, I do think that naming the package `ninjam-server` would actually be *less* confusing:

1. The binary archive downloadable from cockos.com is called `ninjam_server_*.zip`.
2. The link to that archive is labeled "NINJAM server [...]".
3. The directory in the sources is called `ninjam/server`.
4. The name spells out what the package is for more clearly than the abbreviation `ninjamsrv`, IMO.

Maybe we should get some additional opinions on this.

packages/ninjamsrv/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+
+_name=ninjam
+pkgname=ninjamsrv
+pkgver=0.080
+pkgrel=1
+pkgdesc='Ninjam online jamming server'
+arch=(x86_64 aarch64)
+url='https://cockos.com/ninjam/'
+license=(GPL2)
+depends=(gcc-libs)
```
SpotlightKid, line 13: Waiting for more feedback from IRC archlinux-aur on whether having `systemd` in `depends` (probably not) or `optdepends` is useful...

cbix: `depends` doesn't seem to make sense from [it's definition](https://wiki.archlinux.org/title/PKGBUILD#depends). You can run `ninjamsrv` without systemd. Then, `optdepends` IMO doesn't make much sense either because every Arch installation has systemd (see `pacman -Si base`). There are cases where systemd is not installed (non-systemd derivatives, archlinux docker container) but in all of these it's not desired to use systemd for running ninjamsrv.

SpotlightKid: Ok, agreed, that systemd is a special case. Though a package being in `base` apparently doesn't preclude it from being in `depends. See `gcc-libs`.

cbix: TBH even the gcc-libs dependency really only makes sense if it's properly versioned or uses soname depends, so whenever the required gcc-libs version is not there an install would fail 🤔 But that would require the gcc-libs package to properly list all its libprovides :D

SpotlightKid (approved):
Looks good, but I'm confused about the significance of the `license.txt` file in the `nimjam` directory. Does that apply to this package?

cbix:
> Looks good, but I'm confused about the significance of the `license.txt` file in the `nimjam` directory. Does that apply to this package?

I'm quite sure it doesn't. The server directory also includes a `license.txt` citing the GPLv2 and the code itself mentions GPLv2. Even if it's dual licensed like that we can still choose GPL.

PR #84, opened 2022-07-03
update jamulus to 3.9.0

cbix:
currently using the prerelease, change to 3.9.0 when the release is out!

### DIFF
--- a/packages/jamulus/PKGBUILD
+++ b/packages/jamulus/PKGBUILD
@@ -5,31 +5,31 @@
 
 pkgbase=jamulus
 pkgname=(jamulus jamulus-headless)
-pkgver=3.8.2
-pkgrel=5
+pkgver=3.9.0
+pkgrel=1
 pkgdesc="Internet jam session software"
 arch=(x86_64 aarch64)
 url='https://jamulus.io/'
 license=(GPL2 BSD custom:STK)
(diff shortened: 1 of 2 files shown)


packages/jamulus/PKGBUILD
```diff
@@ -5,39 +5,39 @@
...
 build() {
   cd $_pkgsrc
-  qmake "CONFIG+=nosound headless" TARGET=jamulus-headless
+  qmake6 "CONFIG+=serveronly headless" TARGET=jamulus-headless
   make clean
   make
-  qmake "CONFIG+=noupcasename"
+  qmake6 "CONFIG+=noupcasename"
   make clean
   make
```
jujudusud, line 34: I don't if it is good for our usage but, the COMPILING.md file from Jamulus repository says:
`make distclean
qmake6
make`
there is no need to `make clean` after qmake since `make distclean` is done before.

cbix: IIRC there were issues when recompiling, since we're compiling one variant after another

jujudusud: Same here with `make distclean` and using `makepkg`.

jujudusud:
Jamulus 3.9.0 is out !
You can merge this PR if you want.

PR #85, opened 2022-07-04
package vst3sdk

cbix:
- [ ] package static libs so the SDK finds them??
- [ ] prevent "install" on build (by default creates links under `~/.vst3`...)


cbix:
oops, I somehow missed the package is [already in community](https://archlinux.org/packages/community/any/vst3sdk/) 🤦 Will try later to build some plugins using that.

PR #96, opened 2022-07-17
Package some first vcvrack plugins

cbix:
Trying to put together a basic PKGBUILD template for packaging VCV Rack plugins. Many more to come, see #97.

### Why should we package VCV Rack plugins?
- By default, a user needs to sign up for the [VCV Library](https://library.vcvrack.com) in order to download and install 3rd party plugins. This may not be possible or desirable for all users.
- Not all plugins provide pre-built binaries.
- We can easily build plugins for aarch64, which is not officially supported yet by VCV.
- A user can install many plugins by simply typing `pacman -S vcvrack-plugins`.


packages/vcvrack-ahornberg/PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+  patch -p1 -i ../fix-gcc-format.patch
+}
+
+build() {
+  cd $_name-$pkgver
+  # vcvrack make flags
+  export SLUG=$_slug
+  export VERSION=$pkgver
+  export RACK_DIR=/usr/share/vcvrack
+  make dist
```
SpotlightKid, line 33: Can these just be passed as make variables?

    make SLUG=$_slug VERSION=$pkgver RACK_DIR=/usr/share/vcvrack dist

(same in the other `PKGBUILD` files)

packages/vcvrack-sonusmodular/PKGBUILD
```diff
@@ -0,0 +1,32 @@
...
+  export SLUG=$_slug
+  export VERSION=$pkgver
+  export RACK_DIR=/usr/share/vcvrack
+  make dist
+}
+
+package() {
+  cd $_name-$pkgver
+  install -d "$pkgdir"/usr/lib/vcvrack/plugins
+  cp -va dist/$_slug -t "$pkgdir"/usr/lib/vcvrack/plugins
```
SpotlightKid, line 31: Do we need to install:

* the font licenses
* the Windows VST(3) DLLs
* the `LICENSE.txt` file in the plugin dir?

SpotlightKid (changes requested):

packages/vcvrack-ahornberg/PKGBUILD
cbix: Good idea, looks a bit cleaner like that!
a5e987cfef40bc5bff1aeb18cb7cfc833f2f7eba

packages/vcvrack-sonusmodular/PKGBUILD
cbix: Removed the DLLs in 0b27b4cb5b6a55b689989804f1a399f8fd7ad563.

Regarding license files, not sure, the original VCV Library releases include the files in the package ... But maybe we should go and move them to `/usr/share/licenses/$pkgname` or in case of a common license, remove it. I'll think about it.

cbix: I decided to put the font licenses to `/usr/share/licenses/$pkgname` (as they're not common licenses) but discard the GPLv3 license files.

packages/vcvrack-ahornberg/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+_name=Ahornberg-VCV-Modules
+pkgname=vcvrack-ahornberg
+pkgver=2.1.6
+pkgrel=1
+pkgdesc='VCV Rack plugin collection'
+arch=(x86_64 aarch64)
+url='https://github.com/Ahornberg/Ahornberg-VCV-Modules'
+license=(GPL3 custom:OFL)
+groups=(proaudio vcvrack-plugins)
+depends=(gcc-libs vcvrack)
```
SpotlightKid, line 14:
```con
$ find-libdeps  out/vcvrack-ahornberg-2.1.6-1-x86_64.pkg.tar.zst 
libRack.so=libRack.so-64
libOpenGL.so=0-64
libglfw.so=3-64
libjansson.so=4-64
libstdc++.so=6-64
libm.so=6-64
libgcc_s.so=1-64
libc.so=6-64
```

`OpenGL`, ` glfw` and `jansson` are covered by the dependency on `vcvrack, the question is whether these should be considered transitive dependencies or not.

`libRack.so` should be maybe added to the `depends` in the `package` function?

SpotlightKid: This probably needs to be addressed for the other two packages as well.

SpotlightKid (changes requested):

packages/vcvrack-ahornberg/PKGBUILD
cbix: > `OpenGL`, `glfw` and `jansson` are covered by the dependency on `vcvrack`, the question is whether these should be considered transitive dependencies or not.

I'd consider them transitive dependencies, since the plugins themselves only include the `rack.hpp` header and Rack's plugin makefile links these libraries. In fact, my `plugins.patch` for the `vcvrack` package is doing that and I don't remember why all these libs were necessary for building plugins. I should definitely check that, otherwise if Rack eventually changes some dependencies we have to rebuild plugins or otherwise have invalid linked libraries in these plugins. However that would be an ABI-breaking change for Rack and would break a lot of plugins on their side as well...

> `libRack.so` should be maybe added to the `depends` in the `package` function?

Just tried that and it breaks installing the plugin, because the `libRack.so` provided by `vcvrack` is not versioned and makepkg makes the dependency `libRack.so=libRack.so-64` for some reason. I could probably go and make up a versioned soname myself in the `vcvrack` package, but that feels a bit too invasive :/

SpotlightKid: Ok, I guess it's fine then.

SpotlightKid (approved):

PR #119, opened 2022-09-30
feat: import and update package 'ot-simian' from AUR

SpotlightKid:


SpotlightKid:
Note: before merging I want to address the issues marked with `FIXME:` in the PKGBUILD.

SpotlightKid:
CI check currently fails because the build is not reproducible. I have found no indication that the `ldc2` compiler is actually capable of making reproducible builds or whether it supports the `SOURCE_DATE_EPOCH` env var.

The `dmd` compiler supports it, but using `dmd` for building Dplug plugin on linux is not supported :disappointed: 

How should we handle that?

SpotlightKid:
Another question: is it worth packaging `dplug-build` separately?

cbix:
@SpotlightKid if there's no easy way to make it reproducible for now, I'd leave it like that and open an issue here and/or upstream to tackle that.

Btw, [new release](https://punklabs.com/blog/onetrick-simian-1.0.1-released) came out today 🎉

SpotlightKid:
Upstream has no issue tracker :disappointed: 

Time fire up the email client...

cbix:
Right. If it's a Dplug issue however we could report it [there](https://github.com/AuburnSounds/Dplug).

SpotlightKid:
```
To: contact@punklabs.com
Subject: OneTrick Simian reproducible builds?

Hello there,

great plugin and thanks for releasing it under the GPL!

I have created an Arch user repository package for OneTrick SIMIAN, to help users compile it as an LV2 or VST3 format plugin for Arch Linux:

https://aur.archlinux.org/packages/ot-simian

We also want to provide binary packages for Arch Liux users via our third-party binary package repository at https://arch.osamc.de.

We noticed that the produced binaries are not reproducible (as in https://reproducible-builds.org/). On Linux only the ldc2 compiler is supported for compiling Dplug plugins and I could not find out how to make reproducible builds with it.

Do you know whether its is possible to make reproducible builds with Dplug on Linux and, if so, how?

Might I also suggest that you also offer the LV2 version for Linux for sale? As far as I have tested, it works well.


Christopher Arndt
```

SpotlightKid:
Dplug issue: https://github.com/AuburnSounds/Dplug/issues/715

p0nce:
Hello, I'm not sure it's worth it to make a `dplug-build` package.
Note that you can often use a `dplug-build` from a different Dplug version than strictly used to build a plugin, though using the same Dplug to build a plugin and `dplug-build` itself is safest.
Dplug follows Semver, so in this case the latest `dplug-build` from Dplug v12 tag will be able to build any plugin using Dplug v12 even from an earlier version.
`dplug-build` is also not strictly necessary on Linux, you can extract the DUB commands you are interested in with `dplug-build -v` flag.

SpotlightKid:
> dplug-build is also not strictly necessary on Linux, you can extract the DUB commands you are interested in with dplug-build -v flag.

Good to know, thanks!

SpotlightKid:
> > dplug-build is also not strictly necessary on Linux, you can extract the DUB commands you are interested in with dplug-build -v flag.
> 
> Good to know, thanks!

It doesn't show the commands to generate the LV2 `manifest.ttl` file, though.

p0nce:
@SpotlightKid I've been imprecise, `dplug-build` is necessary to build the VST2

SpotlightKid:
> The code that does this is:

Nice. I converted this into Python:

https://github.com/osam-cologne/archlinux-proaudio/blob/feat/package-ot-simian/packages/ot-simian/generate-manifest-ttl.py

cbix (approved):
Didn't test the package yet but PKGBUILD and CI logs look good, nice work here! :)

PR #120, opened 2022-09-30
feat: import package linux-show-player from AUR

SpotlightKid:


SpotlightKid:
Note: before merging, first need to import `python-mido` as well.

SpotlightKid:
I'm only a co-maintainert of the [AUR package](https://aur.archlinux.org/packages/linux-show-player), I can't add `osamc` as a co-maintainer. So sync to AUR is not enabled on purpose and pkgrel is 1, same as in AUR.

If and when a new release comes out, I can contact the AUR maintainer and ask him to add `osamc` or disown the package so I can become maintainer.

nvchecker/archlinux-proaudio.toml
```diff
@@ -80,6 +80,13 @@ github = "olegkapitonov/Kapitonov-Plugins-Pack"
 use_max_tag = true
 include_regex = '^\d+(\.\d+)+$'
 
+[linux-show-player]
+source = "github"
+github = "cbix/ninjam"
```
cbix, line 85: This looks wrong 🤔

packages/linux-show-player/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+  'portmidi: for portmidi support'
+  'python-jack-client: for JACK output support'
+)
+options=('!emptydirs')
+source=("$pkgname-$pkgver.tar.gz::https://github.com/FrancescoCeruti/$pkgname/archive/v$pkgver.tar.gz")
+sha256sums=('e5b70def979984053bae1023210f4c568f989f7714d712506911de6ea3837d0b')
+
+build() {
+  cd $pkgname-$pkgver
+  python setup.py build
```
cbix, line 29: It may be worth mentioning that the [Arch Wiki recommends](https://wiki.archlinux.org/title/Python_package_guidelines#Standards_based_(PEP_517)) packaging python packages using a new method, unless that doesn't work for some reason. I don't have much experience with either but it apparently requires adding two dependencies instead of one :D

cbix (changes requested):
Nice, will try testing this later :)

cbix (approved):

PR #122, opened 2022-10-01
feat: import package a-sid from AUR

SpotlightKid:


packages/a-sid/PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+package() {
+  cd $_reponame-$pkgver
+
+  # plugin bundle
+  install -Dm755 vst3/build/$_reponame.vst3/desktop.ini \
+    vst3/build/$_reponame.vst3/Plugin.ico \
+    -t "$pkgdir"/usr/lib/vst3/$_reponame.vst3
+  install -Dm755 vst3/build/$_reponame.vst3/Contents/Info.plist \
+    vst3/build/$_reponame.vst3/Contents/PkgInfo \
+    -t "$pkgdir"/usr/lib/vst3/$_reponame.vst3/Contents
```
cbix, line 40: These should be 644, or am I missing anything?

cbix (changes requested):

packages/a-sid/PKGBUILD
SpotlightKid: Right. I'll fix it.

Tbh, I'm not sure whether these files even make sense to install on Linux. Maybe some file manager will use the icon for displaying the bundle? Note sure.

The `.plist` file has exe permissions in the source dist, but I surmise that's by mistake.

cbix (approved):

PR #132, opened 2022-10-17
update Jamulus to 3.9.1

jujudusud:
Updating jamulus PKGBUILD to 3.9.1.
Updating jamulus nvchecker to 3.9.1.

PR #133, opened 2022-10-18
update to 3.9.1

SpotlightKid:
Obsoletes: #132

### DIFF
--- a/nvchecker/old_ver.json
+++ b/nvchecker/old_ver.json
@@ -13,7 +13,7 @@
   "ildaeil": "1.2",
   "jack-matchmaker": "0.10.0",
   "jalv-select": "1.3",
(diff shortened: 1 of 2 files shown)


jujudusud:
I have to change the icon path. Let me just a few jours.

SpotlightKid:
This PR already fixes it.

jujudusud:
Nope, the icon should be in "scalable" because it is a svg.

SpotlightKid:
Oh, you mean the install location. Well, it was wrong before and it still works, so no rush. But feel free to make PR with an updated package release.

But tbh, I'd prefer if the `Jamulus.pro` file was fixed to support `PREFIX=/usr`, so that we can use `make install` instead of having custom install commands in the `PKGBUILD` file.

PR #141, opened 2022-10-31
feat: import package 'abcpp' from AUR

SpotlightKid:


packages/abcpp/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+groups=(abc pro-audio)
+source=("https://sourceforge.net/projects/abcplus/files/$pkgname/$pkgname-$pkgver.tar.gz")
+sha256sums=('05a9a7f81b35aff8f1b03ad0c41539508812ec1e52d2f584a180d3c12e35edc1')
+
+prepare() {
+  cd $pkgname-$pkgver
+  sed -i \
+    -e 's|\$(CC) \$(CFLAGS)|$(CC) $(CFLAGS) $(LDFLAGS) -D VERSION="$(VERSION)"|' \
+    -e '/^CFLAGS/d' \
+    Makefile
```
cbix, line 21: Would be nice to have this as a proper patch and submit upstream. If that's not desired for some reason we should probably override these variables in the build step instead :)

cbix (approved):
LGTM!

packages/abcpp/PKGBUILD
SpotlightKid: Upstream has no repo, but I've put the Makefile changes in a patch and will send it via email to the developer.

PR #143, opened 2022-10-31
feat: import package 'littlefly.lv2' from AUR

SpotlightKid:


packages/littlefly.lv2/PKGBUILD
```diff
@@ -0,0 +1,44 @@
...
+# Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
+
+_name=LittleFly
+pkgname=${_name,,}.lv2
+pkgver=1.0
+pkgrel=3
+pkgdesc='A high gain overdrive/distortion pedal simulation LV2 plugin'
+arch=(aarch64 x86_64)
+url="https://github.com/brummer10/${_name}.lv2"
+license=(0BSD)
```
cbix, line 11: This should be changed to BSD, see [this comment](https://gitlab.archlinux.org/pacman/namcap/-/merge_requests/5#note_59140)
```suggestion
license=(BSD)
```

cbix (changes requested):

packages/littlefly.lv2/PKGBUILD
SpotlightKid: I don't follow the argument in the linked comment. IMHO, the 0BSD is fundamentally different license than any of the 1, 2, or whatever-clause BSD licenses, because it has no requirements that must be followed.

https://spdx.org/licenses/0BSD

Stating the license as simply 'BSD' instead of '0BSD' would drop important information, IMO.

cbix: TBH I don't have a strong opinion on whether 0BSD should be covered by BSD or not, but you have a point.

However according to PKGBUILD rules, if a license is not _common_ (i.e. included in the `licenses` package) or _special_ (from a fixed list mentioned [here](https://wiki.archlinux.org/title/PKGBUILD#license), it must be prefixed with `custom:`.

In the long term we should try to reach consensus across packagers though and probably add 0BSD as a special license after all ...

SpotlightKid: Let's try to get a third opinion from `#archlinux-aur`, for example.

cbix (approved):

packages/littlefly.lv2/PKGBUILD
SpotlightKid: Info from IRC:

```
<Foxboron> strogon14: The solution is to standardize on the SPDX license identifiers
<Foxboron> https://spdx.org/licenses/
<phrik> Title: SPDX License List | Software Package Data Exchange (SPDX) (at spdx.org)
<Foxboron> [But] We havent implemented said solution yet
```

So, we'll have to go with `custom:0BSD`  for now.

PR #151, opened 2022-11-06
fix: make sure pacman db is updated when installing openssh

SpotlightKid:


cbix:
This doesn't do any harm here (besides taking extra time), but in general the idea of the shared pacman cache for all steps was that we have an immutable state of upstream repositories across all CI steps. This is really only relevant for the build vs. repro steps to ensure consistent environments so merging this isn't really a problem :)

SpotlightKid:
Ok, noted.

PR #152, opened 2022-11-10
feat: import package 'abcmidi' from AUR

SpotlightKid:


cbix (approved):
Looks good, thanks for this! What confused me is their versioning scheme, apparently it uses major.minor + YYYY-MM-DD but the tags are only the date 🤔

But since the date thing seems to be the more common (and consistent) reference, I agree we should use just that.

SpotlightKid:
Yeah, its a bit wierd. The version number shows up in the help message, but is used nowhere else importnant. And in the interest of keeping the package versioning scheme compatible with the AUR package, I think this is best.

PR #154, opened 2022-11-13
feat: import and update package 'triceratops' from AUR

SpotlightKid:


packages/triceratops/PKGBUILD
```diff
@@ -0,0 +1,54 @@
...
+
+check() {
+  cd $pkgname
+  mkdir -p lv2
+  ln -sf ../build/$pkgname.lv2 lv2/$pkgname.lv2
+  LV2_PATH="$PWD/lv2:/usr/lib/lv2" lv2lint \
+    -s _Z* \
+    -s _bss* \
+    -s __bss* \
+    -s __end__ \
```
SpotlightKid, line 45: Any idea how to get `waf` to link the plugin libs, so that these symbols are not exported?

Probably default `LDFLAGS` from `makepkg` are not taken into account. How to get `waf` to do that?

SpotlightKid: Got it: `-fvisibility=hidden` needs to be added to `CXXFLAGS`.

nvchecker/archlinux-proaudio.toml
```diff
@@ -263,6 +263,12 @@ github = "swh/lv2"
 use_max_tag = true
 prefix = "v"
 
+[triceratops]
+source = "regex"
+url = "https://sourceforge.net/projects/triceratops/files/"
```
cbix, line 268: Cool if this works reliably, but did you try using the rss similar to the tuxguitar section below?

packages/triceratops/PKGBUILD
```diff
@@ -0,0 +1,68 @@
...
+makedepends=(python lv2)
+checkdepends=(lv2lint)
+optdepends=('lv2-host: for loading the LV2 plugin')
+provides=($pkgname.lv2)
+conflicts=($pkgname.lv2)
+source=("https://downloads.sourceforge.net/project/triceratops/${pkgname}_v${pkgver}_Oct_2022.tar.gz"
+        'triceratops-gui-instantiate.patch'
+        'triceratops-remove-printf.patch'
+        'triceratops-symbol-visibility.patch'
+        'triceratops-ttl.patch')
```
cbix, line 24: If these patches are identical to the PRs, you can also source them directly using `triceratops-gui-instantiate.patch::https://patch-diff.githubusercontent.com/raw/thunderox/triceratops/pull/11.patch`

SpotlightKid: Unfortunately, since I created these the PRs with these patches independently, they contain conflicting whitespace changes and can not be used as is.

cbix (approved):

PR #155, opened 2022-11-14
feat: import package 'dxconvert' from AUR

SpotlightKid:


SpotlightKid:
Note: I have not added the package to `aur/packages`, because, even though I am the maintainer of [dxconvert](https://aur.archlinux.org/packages/dxconvert), the latter has an optdepends on `castools`, which only exists as [castools-git](https://aur.archlinux.org/packages/castools-git) in the AUR, thus we can't add that dependency in our package, even if it's only optional.

See also the comment in the `PKGBUILD`.

diovudau (approved):

PR #156, opened 2022-11-17
feat: import package 'rezonateur' from AUR

SpotlightKid:


cbix (approved):

cbix:
Namcap reports
```
rezonateur E: Dependency gcc-libs detected and not included (libraries ['usr/lib/libstdc++.so.6', 'usr/lib/libgcc_s.so.1'] needed in files ['usr/lib/vst/rezonateur-vst.so'])
```

SpotlightKid:
Hmm, it reports it as already satisfied here?! Anyway, I threw it in.

SpotlightKid:
Unfortunately, I just discovered that the aarch64 build fails, because the bundled DPF version is too old. Patching now...

PR #164, opened 2022-12-05
Package audiowmark

diovudau:
Please checks for cracks

### DIFF
--- a/aur/packages
+++ b/aur/packages
@@ -2,6 +2,7 @@ abcm2ps
 abcpp
 aether.lv2
 a-sid
+audiowmark
 clap
 darkice
 die-plugins
@@ -55,5 +56,5 @@ vcvrack-freesurface
 vcvrack-goodsheperd
 vcvrack-sonusmodular
 vcvrack-zetacarinae
-ykchorus
 xtuner
+ykchorus
--- a/nvchecker/archlinux-proaudio.toml
+++ b/nvchecker/archlinux-proaudio.toml
@@ -37,7 +37,10 @@ use_max_tag = true
 # note: we use the stable website downloads for building, but gitlab for checking
 [audiowmark]
 source = "gitlab"
+host = "code.uplex.de"
 gitlab = "stefan/audiowmark"
+# Regex to avoid the "proof-of-concept" tag. Start with a number.
+include_regex = '^\d.+$'
 use_max_tag = true
 
 [clap]
--- a/nvchecker/archlinux-proaudio.toml
+++ b/nvchecker/archlinux-proaudio.toml
@@ -34,6 +34,12 @@ source = "github"
 github = "sdangelo/asid"
 use_max_tag = true
 
+# note: we use the stable website downloads for building, but gitlab for checking
+[audiowmark]
+source = "gitlab"
+gitlab = "stefan/audiowmark"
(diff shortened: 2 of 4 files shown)


nvchecker/archlinux-proaudio.toml
```diff
@@ -34,6 +34,12 @@ source = "github"
 github = "sdangelo/asid"
 use_max_tag = true
 
+# note: we use the stable website downloads for building, but gitlab for checking
+[audiowmark]
+source = "gitlab"
+gitlab = "stefan/audiowmark"
+use_max_tag = true
```
cbix, line 41: this needs `host = "code.uplex.de"` and probably an `include_regex` because there is one non-version tag in that repo

cbix (changes requested):

aur/packages
```diff
@@ -2,6 +2,7 @@ abcm2ps
 abcpp
 aether.lv2
 a-sid
+audiowmark
```
SpotlightKid, line 5: Since there is no `audiowmark` package in AUR yet, I'm not sure what will happen, when this is merged. Maybe @cbix can clarify.

packages/audiowmark/PKGBUILD
```diff
@@ -0,0 +1,32 @@
...
+
+pkgname=audiowmark
+pkgver=0.6.1
+pkgrel=1
+pkgdesc='Audio and video watermarking'
+arch=(x86_64 aarch64)
+url='https://uplex.de/audiowmark/'
+license=(GPL3)
+#depends=(fftw libsndfile libgcrypt zita-resampler mpg123 ffmpeg) #explicit from README
+depends=(zita-resampler fftw ffmpeg mpg123) #as namcap sees them
```
SpotlightKid, line 12: When in doubt, follow `namacp`. at least where "normal" dependencies are concerned (don't trust it for Python package dependencies, though).

packages/audiowmark/PKGBUILD
```diff
@@ -0,0 +1,32 @@
...
+sha256sums=('0b22c4364e111f6a70dac0118a1f67319bce25e57c501d5e55918cdb0c2577fb')
+
+build() {
+  cd $pkgname-$pkgver
+  ./configure --prefix=/usr --with-ffmpeg
+  make
+}
+
+check() {
+  #make in build() creates tests/test-common.sh
```
SpotlightKid, line 24: Is that comment important for understanding the PKGBUILD?

packages/audiowmark/PKGBUILD
```diff
@@ -0,0 +1,32 @@
...
+  cd $pkgname-$pkgver
+  make check
+}
+
+package() {
+  cd $pkgname-$pkgver
+  make prefix="$pkgdir"/usr install
```
SpotlightKid, line 31: I would suggest installing `README.adoc` as documentation to `/usr/share/doc/$pkgname`, since there is no man page and `audiowmark --help` only gives a lost of options.

SpotlightKid (changes requested):

aur/packages
cbix: It will simply be added by the `osamc` user.

cbix (approved):

PR #168, opened 2022-12-08
feat: add package 'python-miditk-smf'

SpotlightKid:
Had this lying around for a while. No reason, not to add it, imho.


thisven:
Installs just fine. Good job!

thisven (approved):

PR #172, opened 2022-12-20
feat: update package 'faustlive' to upstream version 2.5.13

SpotlightKid:


cbix:
Looks good, not sure if we should wait for @dvzrv to package the corresponding new faust version 2.54.9 and if FL is actually using the examples/libs from the faust package. I'd guess it's fine to already push the new FaustLive since it's just using faust's shared libs ...

SpotlightKid:
I'm fine with waiting a bit.

PR #175, opened 2022-12-23
package clap-validator

cbix:
[clap-validator](https://github.com/free-audio/clap-validator)

In the future we can use this to `check()` CLAP plugins


SpotlightKid (approved):
Looks good to me.

PR #176, opened 2022-12-23
package clap-example-host

cbix:
[clap-host](https://github.com/free-audio/clap-host) is an example CLAP plugin host (similar to jalv for LV2)


packages/clap-host/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+arch=(aarch64 x86_64)
+url='https://github.com/free-audio/clap-host'
+license=(MIT)
+depends=(gcc-libs qt6-base)
+makedepends=(catch2 cmake rtaudio rtmidi)
+groups=(pro-audio)
+_clap_ref='27b7af0'
+_clap_helpers_ref='4d4e297'
+source=("$pkgname-$pkgver.tar.gz::https://github.com/free-audio/$pkgname/archive/refs/tags/$pkgver.tar.gz"
+        "clap-$_clap_ref.tar.gz::https://github.com/free-audio/clap/archive/$_clap_ref.tar.gz"
```
SpotlightKid, line 17: Can or should this not be compiled against the system installed `clap` package? Or is this tied to a specific version of CLAP?

packages/clap-host/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+  bsdtar -xf ../clap-helpers-$_clap_helpers_ref.tar.gz -C clap-helpers --strip-components 1
+}
+
+build() {
+  cmake -B build -S $pkgname-$pkgver \
+    -DCMAKE_BUILD_TYPE=None \
+    -DCMAKE_INSTALL_PREFIX=/usr \
+    -DCLAP_BUILD_TESTS=On \
+    -Wno-dev
+  cmake --build build --target clap-host clap-tests
```
SpotlightKid, line 36: Better include `$pkgname` in the build directory name.

packages/clap-host/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+}
+
+check() {
+  ctest --test-dir build --output-on-failure
+}
+
+package() {
+  depends+=(hicolor-icon-theme librtaudio.so librtmidi.so)
```
SpotlightKid, line 44: Why not put `hicolor-icon-theme` in the `depends` list in the header?

SpotlightKid (changes requested):

packages/clap-host/PKGBUILD
```diff
@@ -0,0 +1,48 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+pkgname=clap-host
```
SpotlightKid, line 4: I think we should talk with the Arch maintainers about that package name.

packages/clap-host/PKGBUILD
cbix: It's not required at build time, only when installing.

packages/clap-host/PKGBUILD
cbix: Compilation seems to work with clap 1.1.3. However clap-host's CMakeLists.txt adds the clap dependency using `add_subdirectory`, expecting the submodule instead of searching global CMake modules.

We could probably patch that but I'm not that experienced with CMake :/

packages/clap-host/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+pkgver=1.0.2
+pkgrel=1
+pkgdesc='CLAP example host'
```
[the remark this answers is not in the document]
cbix: 095898f04999233fac9f1fce0d6c2dbd33349515

packages/clap-host/PKGBUILD
cbix: Is that sort of a best practice? I usually just use `build` according to the [package guidelines](https://wiki.archlinux.org/title/CMake_package_guidelines#Template). Since we only build one cmake project (the package itself) and the `build` directory is already under `src` it feels reduntant to include the package name but maybe that's just my perspective 🤔

packages/clap-host/PKGBUILD
cbix: @dvzrv for context: this package installs a `clap-host` binary and our concern is that since other hosts have `provides=(clap-host)` this may conflict with `provides` either being a notion of compatibility (package provides a certain shared lib or binary) or of an abstract service implementation (package supports a plugin format).

An idea would be to rename that to `clap-example-host` now through a patch.

packages/clap-host/PKGBUILD
SpotlightKid: It prevents clashing directory names where `srcdir` is shared. That could even maybe happen in our build system if, for example, several packages are build in the same CI run. Or maybe not, but it's easy to prevent if build dirsare unique to packages.

packages/clap-host/PKGBUILD
SpotlightKid: Ok, then I guess we stick to including clap in he source, even though upstream should then really package it in the source archive.

packages/clap-host/PKGBUILD
SpotlightKid: Ok, I see. But I wonder why I never see t done that way in the official repos.

One disadvantage is that wrong dependencies will only be caught when actually installing the package, or is it not?

cbix: `hicolor-icon-theme` is a special package as it only includes the necessary directory structure for the hicolor icon theme, so desktop managers can use that tree to look for icons. Lots of packages install their application icon in that tree, so in order to make the icon "discoverable", they need `hicolor-icon-theme` as a dependency. Also, namcap has a check for that :D

packages/clap-host/PKGBUILD
cbix: Thanks for the clarification. Shared srcdir is a valid case indeed. Changed now in eef7fe0ac66144c5b3f49b2ed0173c0a4c51991e

packages/clap-host/PKGBUILD
dvzrv: Yeah, ideally name the pkg clap-example-host. Pacman unfortunately has a (not very obvious) preference for "real packages" over "virtual provides" (a package named "clap-host" would be auto-chosen over any of the virtual provides).

cbix: Done in 57b9a889f109b48b6f24b08ece92b3c95b873987

SpotlightKid (approved):

PR #182, opened 2022-12-31
package liblc3

cbix:
[liblc3](https://github.com/google/liblc3)

### DIFF
--- a/packages/liblc3/PKGBUILD
+++ b/packages/liblc3/PKGBUILD
@@ -8,7 +8,6 @@ pkgdesc='Low Complexity Communication Codec library and tools'
 arch=(aarch64 x86_64)
 url='https://github.com/google/liblc3/'
 license=(Apache)
-groups=(pro-audio)
 depends=(glibc)
 makedepends=(meson)
 provides=(liblc3.so)
--- a/packages/liblc3/PKGBUILD
+++ b/packages/liblc3/PKGBUILD
@@ -33,4 +33,6 @@ check() {
 
 package() {
   meson install -C build-$pkgname --destdir "$pkgdir"
+  cd $pkgname-$pkgver
+  install -vDm644 README.md -t "$pkgdir"/usr/share/doc/$pkgname
 }
--- a/packages/liblc3/PKGBUILD
+++ b/packages/liblc3/PKGBUILD
@@ -1,0 +1,37 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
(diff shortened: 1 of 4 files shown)


packages/liblc3/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+pkgname=liblc3
+pkgver=1.0.1
+pkgrel=1
+pkgdesc='Low Complexity Communication Codec library and tools'
+arch=(aarch64 x86_64)
+url='https://github.com/google/liblc3/'
+license=(Apache)
+groups=(pro-audio)
```
cbix, line 11: We may actually not add any group here since it's not purely a pro-audio package

packages/liblc3/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+depends=(glibc)
+makedepends=(meson)
+provides=(liblc3.so)
```
SpotlightKid, line 14: Shouldn't this go into the `package` function? I'm not sure if it makes a difference technically, but I think it's better to be consistent, isn't it?

packages/liblc3/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+  arch-meson $pkgname-$pkgver build-$pkgname -D tools=true
+  meson compile -C build-$pkgname
+}
+
+check() {
+  meson test -C build-$pkgname --print-errorlogs
+}
+
+package() {
+  meson install -C build-$pkgname --destdir "$pkgdir"
```
SpotlightKid, line 36: I think the `README.md` should be installed as documentation.

SpotlightKid (changes requested):
Are there any man pages for the library or the tools installed?

packages/liblc3/PKGBUILD
cbix: I've only ever seen that (`provides` inside `package`) for split packages, otherwise it's always set globally.

packages/liblc3/PKGBUILD
cbix: > Are there any man pages for the library or the tools installed?

No, but you can run `elc3 -h` or `dlc3 -h`.

> I think the `README.md` should be installed as documentation.

done: 6be4df78951c503cd18081115e0a2e07267c880c

packages/liblc3/PKGBUILD
SpotlightKid: I thought I saw that with provides for libs somewhere, but now I can't find it. Never mind then.

SpotlightKid (approved):

PR #191, opened 2023-01-21
fix: bump pkgrel of python-soundfile to trigger rebuild

SpotlightKid:
… with latest libsndfile with MP3 support


SpotlightKid:
Note: this update is mainly for the benefit of the AUR version, so it will trigger a rebuild/reinstall for users there.

PR #214, opened 2023-03-16
feat: add new package 'convertwithmoss'

SpotlightKid:


packages/convertwithmoss/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+
+_name=ConvertWithMoss
+pkgname=${_name,,}
+pkgver=6.2.0
+pkgrel=1
+pkgdesc='A tool for converting multi-sample from one format to another'
+url='https://mossgrabers.de/Software/ConvertWithMoss/ConvertWithMoss.html'
+arch=(any)
+license=(LGPL)
+depends=('java-runtime>=16' java-openjfx)
```
cbix, line 13: The package doesn't install (and build) on aarch64 because there is no `java-openjfx` package, should we just set arch to `x86_64` then?

cbix (changes requested):
Nice to see it suddenly being reproducible :D

packages/convertwithmoss/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+
+build() {
+  cd $_name-$pkgver
+  mvn -Dproject.build.outputTimestamp=$SOURCE_DATE_EPOCH clean install
+}
+
+package() {
+  cd $_name-$pkgver
+  install -Dm644 target/lib/*.jar -t "$pkgdir"/usr/share/$_name
+  ln -sf $pkgname-$pkgver.jar "$pkgdir"/usr/share/$_name/$pkgname.jar
```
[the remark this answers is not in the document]
SpotlightKid: The shell wrapper provided in the sources and installed as `/usr/bin/convertwithmoss` assumes the directory is called `<prefix>/share/ConvertWithMoss`. That would need to be patched then. Is that worth it?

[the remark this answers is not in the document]
cbix: Well, nevermind then :) Seems to just be a convention anyway, from the [wiki](https://wiki.archlinux.org/title/Arch_package_guidelines#Directories):
> Use `/etc/pkg` where `pkg` is the name of the package (or a suitable alternative, eg, apache uses `/etc/httpd/`).

packages/convertwithmoss/PKGBUILD
SpotlightKid: Turns out, it isn't needed at all, an architecture-specific package of javafx is apparently downloaded during build and included in the package.

cbix: I see, thanks for looking into that!

cbix (approved):

PR #226, opened 2023-05-06
fix: add jack to checkdepends for package qjackcapture

SpotlightKid:


SpotlightKid:
The problem is the non-functional pyjacklib not a missing checkdepends. -> closing.

Issue #3, opened 2022-02-18
Wishlist for packages


diovudau:
Here are wishes and ideas of what to software to package. 
Inclusion in this list does not mean any obligation or timeline of a package actually happening.
There is no order or priority to this list. 
Entries must not be already in the official arch binary repositories. 

Many of the programs here may have upstream problem that prevent proper packaging, hence they are not already binary packages. In this case they are either "will not happen" or need communication and fixes with upstream.

- tuxguitar https://sourceforge.net/projects/tuxguitar/
- ly2video https://github.com/aspiers/ly2video
- synthclone https://github.com/surfacepatterns/synthclone
- https://github.com/danielappelt/k-board-editor  :)
- https://bitbucket.org/agraef/pizmidi/src/master/pizmidi/midiPolyphony  (Good to make monophonic synths into polyphonic ones through multi-instances)
- klick  https://github.com/dsacre/klick

SpotlightKid:
midiPolyphony is part of the [pizmidi](https://aur.archlinux.org/packages/pizmidi-git) collection, which needs the VST2 headers to compile. These headers are only available in the AUR right now from the package [vst2sdk](web.archive.org/), which builds from an old copy of the VST3 SDK, which still has the VST 2 headers. It downloads this old copy from web.archive.org/. This is reliable.

Apart from that. I don't think we are allowed to distribute binaries compiled using these headers if we're not licensees of Steinberg,

diovudau:
A personal list of software for "review later".  I figure it is more useful to have them here instead in my private offline notebook.
I mostly read the first paragraph of their github readme and glanced at the "LICENSE".

- https://github.com/manucontrovento/SpinTool "SpinTool is a loop based software fully controllable with any midi device, synced with Jack transport. "
- https://github.com/markusschwenk/midieditor   "MidiEditor is a free software providing an interface to edit, record, and play Midi data."
- https://github.com/ElvishArtisan/GlassCoder "GlassCoder is a minimalist audio encoder for IceCast and Shoutcast streaming servers as well as HTTP Live Streaming (HLS)."
-https://github.com/cbdevnet/midimonster  "the MIDIMonster is a universal control and translation tool for"
- https://github.com/magnetophon/VoiceOfFaust " Turns any monophonic sound into a synthesizer, preserving the pitch and spectral dynamics of the input. "

cbix:
[...]

PR #237, opened 2023-05-20
Package simplescreenrecorder

thisven:
As requested by @diovudau on the [Wishlist](https://github.com/osam-cologne/archlinux-proaudio/issues/3#issuecomment-1554416021), I packaged [SimpleScreenRecorder](https://www.maartenbaert.be/simplescreenrecorder) to be maintained here.

### DIFF
--- a/aur/packages
+++ b/aur/packages
@@ -39,6 +39,7 @@ rakarrack-plus
 rezonateur
 rtcqs
 sfzlint
+simplescreenrecorder
 stone-phaser
 string-machine
 swh-plugins.lv2
--- a/aur/packages
+++ b/aur/packages
@@ -39,7 +39,6 @@ rakarrack-plus
 rezonateur
 rtcqs
 sfzlint
-simplescreenrecorder
 stone-phaser
 string-machine
 swh-plugins.lv2
--- a/packages/simplescreenrecorder/PKGBUILD
+++ b/packages/simplescreenrecorder/PKGBUILD
@@ -14,8 +14,8 @@ depends=(alsa-lib desktop-file-utils ffmpeg glu gtk-update-icon-cache jack
          libgl libpulse libx11 libxext libxfixes libxi libxinerama qt5-base
          qt5-x11extras)
 makedepends=(git cmake qt5-tools)
-optdepends=('lib32-simplescreenrecorder: OpenGL recording of 32-bit applications
-            ')
+#optdepends=('lib32-simplescreenrecorder: OpenGL recording of 32-bit applications
+            ') # unused, see: https://github.com/osam-cologne/archlinux-proaudio/pull/237#discussion_r1213428473
 source=("https://github.com/MaartenBaert/$_name/archive/refs/tags/$pkgver.tar.gz"
          ffmpeg5.patch::https://patch-diff.githubusercontent.com/raw/MaartenBaert/ssr/pull/934.patch)
 sha256sums=('69fccf8df6e7285d4df5033eed8748f4eae13b407488906dcf4810a277d557b8'
--- a/packages/simplescreenrecorder/PKGBUILD
+++ b/packages/simplescreenrecorder/PKGBUILD
@@ -14,8 +14,6 @@ depends=(alsa-lib desktop-file-utils ffmpeg glu gtk-update-icon-cache jack
          libgl libpulse libx11 libxext libxfixes libxi libxinerama qt5-base
          qt5-x11extras)
 makedepends=(git cmake qt5-tools)
-#optdepends=('lib32-simplescreenrecorder: OpenGL recording of 32-bit applications
-            ') # unused, see: https://github.com/osam-cologne/archlinux-proaudio/pull/237#discussion_r1213428473
 source=("https://github.com/MaartenBaert/$_name/archive/refs/tags/$pkgver.tar.gz"
          ffmpeg5.patch::https://patch-diff.githubusercontent.com/raw/MaartenBaert/ssr/pull/934.patch)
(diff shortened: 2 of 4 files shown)


aur/packages
```diff
@@ -39,6 +39,7 @@ rakarrack-plus
 rezonateur
 rtcqs
 sfzlint
+simplescreenrecorder
```
SpotlightKid, line 42: We can only update packages in the AUR, for which we can ad the `osamc` account as a (co-)maintainer, which is not the case here. This line should be removed.

packages/simplescreenrecorder/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+
+_name=ssr
+pkgname=simplescreenrecorder
+pkgver=0.4.4
+pkgrel=1
+pkgdesc='A feature-rich screen recorder that supports X11 and OpenGL'
+arch=(x86_64 aarch64)
+url='https://www.maartenbaert.be/simplescreenrecorder'
+license=(GPL3)
+groups=(pro-audio)
```
SpotlightKid, line 12: I wouldn't classify ssr as audio software, so I don't think the group fits here and should be removed.

packages/simplescreenrecorder/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+depends=(alsa-lib desktop-file-utils ffmpeg glu gtk-update-icon-cache jack
+         libgl libpulse libx11 libxext libxfixes libxi libxinerama qt5-base
+         qt5-x11extras)
+makedepends=(git cmake qt5-tools)
+optdepends=('lib32-simplescreenrecorder: OpenGL recording of 32-bit applications
```
SpotlightKid, line 17: We shouldn't depend on AUR packages, even optionally, in packages in our repo. Users should be able to use our repo without using the AUR, IMHO.

packages/simplescreenrecorder/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+            ')
+source=("https://github.com/MaartenBaert/$_name/archive/refs/tags/$pkgver.tar.gz"
+         ffmpeg5.patch::https://patch-diff.githubusercontent.com/raw/MaartenBaert/ssr/pull/934.patch)
+sha256sums=('69fccf8df6e7285d4df5033eed8748f4eae13b407488906dcf4810a277d557b8'
+            '4b01938615a34127236a21ee0ffa20bbb179c8bfcc5ecef872fc5a246727d2e1')
+
+prepare() {
+  cd $_name-$pkgver
+  mkdir -p build
+  patch -p1 -i ../ffmpeg5.patch # Fix build with ffmpeg 5
```
SpotlightKid, line 27: Shouldn't this be `"$srcdir"/ffmpeg5.patch`?

packages/simplescreenrecorder/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+}
+
+build() {
+  cd $_name-$pkgver/build
+  cmake -DCMAKE_INSTALL_PREFIX="/usr" -DCMAKE_BUILD_TYPE=Release \
+    -DWITH_QT5=on \
+    -DCMAKE_INSTALL_LIBDIR='lib' ../
```
SpotlightKid, line 34: Minor niggle: format command with one `-D` option per line.

SpotlightKid (changes requested):

packages/simplescreenrecorder/PKGBUILD
thisven: And what if we build that package, too?

SpotlightKid: As discussed in the Matrix chat, that would require enabling the `multilib` repo in our CI.

Let's wait for the input from @cbix on that.

cbix: I wouldn't add any packages to our main proaudio repo that depend repos other than core/extra/community. If there is some need for it, we may add a separate `proaudio-multilib` repo. Also, recording OpenGL seems to still be an experimental feature and I'd say it's not worth the effort just to support this super tiny edge case :)

thisven: I see. I put a comment before the `optdepends` line and provide the link to this conversation for transparency on this decision.

thisven: OK, that doesn't seem to work very well in the CI. I just remove the line.

packages/simplescreenrecorder/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+pkgver=0.4.4
+pkgrel=1
+pkgdesc='A feature-rich screen recorder that supports X11 and OpenGL'
+arch=(x86_64 aarch64)
+url='https://www.maartenbaert.be/simplescreenrecorder'
+license=(GPL3)
+depends=(alsa-lib desktop-file-utils ffmpeg glu gtk-update-icon-cache jack
+         libgl libpulse libx11 libxext libxfixes libxi libxinerama qt5-base
+         qt5-x11extras)
+makedepends=(git cmake qt5-tools)
```
SpotlightKid, line 16: Is `git` still needed to build?

thisven: I don't think so. Let's see...

SpotlightKid (approved):
LGTM. Be sure to squash the commits of the PR when merging (the default).

packages/simplescreenrecorder/PKGBUILD
thisven: It builds like a charm.

PR #239, opened 2023-05-24
feat: update package faustlive to 2.5.15-1

SpotlightKid:


cbix (approved):
Didn't test since I'm still on vacation but looks good 👍

PR #254, opened 2023-07-20
feat:  send CI notifications to sonoj.org matrix server

SpotlightKid:
Use [custom drone.io plugin](https://github.com/SpotlightKid/drone-matrixchat-notify) to send CI notifications to sonoj.org matrix server

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -441,3 +441,45 @@ when:
   status:
     - success
     - failure
+
+---
+kind: pipeline
+name: Send notification
+
+steps:
+- name: Notify Matrix chat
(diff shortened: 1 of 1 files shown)


SpotlightKid:
Alas, the drone.io matrix plugin seems to be broken:

https://github.com/drone-plugins/drone-matrix/issues/32

cbix (approved):
Nice work! Not sure if notifying on PRs is of that much value, but let's try this for a while 👌

SpotlightKid:
I want this mainly for builds which take a very long time, so I don't need to check constantly whether they have already finished.

PR #265, opened 2023-08-25
[ci] fix: update package database when upgrading packages

SpotlightKid:
Also:

* Don't suppress pacman output.
* Install pipeline step deps in step's command list, not hidden in script.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -27,7 +27,7 @@ steps:
       - name: pkgdb
         path: /var/lib/pacman/sync
     commands:
-      - pacman -S --noconfirm namcap >/dev/null
+      - pacman -S --noconfirm namcap
       - if [ -n "$PACKAGE" ]; then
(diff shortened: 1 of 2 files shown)


cbix:
@SpotlightKid just came across this and #267 after [this ci failure](https://ci.cbix.de/osam-cologne/archlinux-proaudio/1813/1/9) confused me – this failed because the "build" and "reproduce" step were operating on different package versions. Previously the "prepare" step would perform a package db update (`-Sy`) and the other steps are mounting the pacman db + cache as a volume, to guarantee a consistent pacman db throughout the pipeline.

Was the issue you were fixing here somehow related to that or can we simply change the `-Syu` back to `-Su`?

SpotlightKid:
It's been a while, so I don't remember exactly, but I think here was an issue with signing keys not updating properly and so the second update would fail.

PR #288, opened 2023-10-16
ola: use legacy protobuf to fix build

### LINKED ISSUE #251: installing protobuf (23.4-1) breaks dependency 'libprotobuf.so=32-64' required by ola
Dear maintainer,

I get

```
:: installing protobuf (23.4-1) breaks dependency 'libprotobuf.so=32-64' required by ola
```

during `# pacman --sync --refresh --sysupgrade` and found a [similar issue in this repo a year ago](https://github.com/osam-cologne/archlinux-proaudio/pull/61). I don't understand what made the difference there except of a rebuild, but:

Can you fix the current issue in this repo, too?

cbix:
- Closes #251 
- blocked by #330
- (actually) blocked by OpenLightingProject/ola#1879

### DIFF
--- a/packages/ola/.gitignore
+++ b/packages/ola/.gitignore
@@ -1,0 +1,1 @@
+/fix-protobuf-v20.patch
--- a/packages/ola/PKGBUILD
+++ b/packages/ola/PKGBUILD
@@ -10,7 +10,7 @@ arch=(x86_64 aarch64)
 url='https://www.openlighting.org'
 license=(LGPL2.1 GPL2)
 depends=()
-makedepends=(avahi cppunit doxygen flake8 libftdi-compat liblo libmicrohttpd ncurses protobuf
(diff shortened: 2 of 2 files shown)


StephanRichter:
Tried this patch. Now the build goes ahead of `checking protoc version...`
and fails with a new error:

```bash
checking for google/protobuf/compiler/command_line_interface.h... no
configure: error: Cannot find the protoc header files
```

topas-rec:
Good idea for the workaround and thanks for thinking about it and working this out.

I'll try to build that soon™.

vermaete:
Yocto / Openembedded is now at version 23.4.0 of protoc.
Besides going back in time, is there already another solution?
Switching back in protoc version is rather complicated (or not done) in Yocto.

So the Yocto release 'Mickeldore' is still fine.  The next release 'Nanbield' and master isn't anymore.

PR #322, opened 2023-12-04
vcvrack: vendor rtaudio

cbix:
The new rtaudio 6 is not installable with the current vcvrack package due to its dependency on the older soname.
Since Rack uses some random patched version of rtaudio and we're nowadays not fine anymore with a simple patch, I'm vendoring the library again (as intended by the author).

I'm also now using git sources to simplify packaging.


packages/vcvrack/PKGBUILD
```diff
@@ -88,31 +83,31 @@ prepare() {
...
   cd ../$_plugin_name-$_plugin_ver
-  VERSION=$_plugin_ver RACK_DIR=../$_name-$pkgver \
+  VERSION=$_plugin_ver RACK_DIR=../$_name \
     LDFLAGS+=" $(pkg-config --libs samplerate)" make dist
 }
 
 package() {
   # Rack does not start with glfw-wayland
-  depends+=(libcurl.so libGLEW.so libarchive.so libcrypto.so librtaudio.so librtmidi.so
+  depends+=(libcurl.so libGLEW.so libarchive.so libcrypto.so librtmidi.so
```
cbix, line 108: ah, I forgot to add transitive dependencies from rtaudio, i.e. alsa/jack/pulse. Wondering though if all these are hard requirements for Rack to work (as suggested by the rtaudio package) or if they're actually optional, so it works with either of the 3.

PR #330, opened 2023-12-11
Package protobuf-21

cbix:
- This is required for #251
- Not adding to nvchecker because the whole point is to pin a specific version
- Blocked by #332 

TODO
- [x] improve packaging so it can be used as a makedep
- [x] improve packaging so it doesn't conflict with protobuf
- [x] testing with #288 
  - apparently for building ola we'd also need an older `abseil-cpp`. Installing/running the existing ola package works with this package installed.


packages/protobuf21/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+    -D CMAKE_BUILD_TYPE=None
+    -D CMAKE_INSTALL_PREFIX=/usr
+    -D CMAKE_INSTALL_INCLUDEDIR=/usr/include/$pkgname
+    -D protobuf_BUILD_SHARED_LIBS=ON
+    -D protobuf_BUILD_TESTS=OFF
+    -S $_name-$pkgver
+    -W no-dev
+  )
+
+  cmake "${cmake_options[@]}"
```
SpotlightKid, line 51: I have never understood what is the advantage of putting the cmake options in an array of specifying them directly. They are only used once, after all.

But this is just a comment, no need to change it.

packages/protobuf21/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+  cmake --build build --verbose
+
+  ln -rs build $_name-$pkgver/src/.libs
+  cd $_name-$pkgver/python
+  PROTOC="$srcdir"/build/protoc \
+    python -m build --wheel --no-isolation
+}
+
+package_protobuf21() {
```
SpotlightKid, line 60: This isn't a split package, isn't it?

Just name the function `package`, please.

packages/protobuf21/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+  provides=("protobuf=$pkgver" libprotoc.so libprotobuf.so libprotobuf-lite.so)
+  DESTDIR="$pkgdir" cmake --install build
+  cd "$pkgdir"
+  mv usr/bin/protoc{,21}
+  rm -r usr/lib/{cmake,pkgconfig,*.so}
```
SpotlightKid, line 65: How does software wanting to compile against this version of the lib find it without the pkgconf file?

Will it rely on the pkgconf files from the package for the current version?

packages/protobuf21/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+
+_name=protobuf
+pkgname=protobuf21
+# DO NOT UPDATE, the point of this package is to pin the version
+pkgver=21.12
+pkgrel=1
+pkgdesc='Protocol Buffers (legacy)'
+arch=(aarch64 x86_64)
+url='https://developers.google.com/protocol-buffers/'
+license=(BSD)
```
SpotlightKid, line 21: namcap complains:

    protobuf21 E: Missing custom license directory (usr/share/licenses/protobuf21)

SpotlightKid (changes requested):

packages/protobuf21/PKGBUILD
cbix: IMO the main benefit is that you don't need `\` at the end of each line

packages/protobuf21/PKGBUILD
cbix: I previously just moved that to `usr/lib/protobuf21/pkgconfig`, but wasn't sure how to test if it works. For ola (the reason we need this package in the first place) they just check for some header files, anyway building doesn't work unless we also package an earlier `abseil-cpp` :/

Guess I'll go with renaming it again :)

SpotlightKid: I'm not sure renaming would be sufficient, you probably also need to change the line

    Libs: -L${libdir} -lprotobuf

in `protobuf.pc` to pass the correct version for the `-l` option. Maybe have  look how other compatibility packages do it?

But if this package is only used for compiling `ola`, I'm not sure it's worth the effort. I just want to make sure that this package does not depend on the normal `protobuf`  package.

cbix: One example I found was [podofo-0.9](https://gitlab.archlinux.org/archlinux/packaging/packages/podofo-0.9/-/blob/main/PKGBUILD), which is [used by scribus](https://gitlab.archlinux.org/archlinux/packaging/packages/scribus/-/blob/main/PKGBUILD) by passing the include/lib paths during build, even though there is a pkgconfig under `/usr/lib/podofo-0.9/pkgconfig`.

Well, also most "legacy" packages I found provide a custom libdir and link the versioned `.so` to `/usr/lib`. I'll do that again here.

packages/protobuf21/PKGBUILD
cbix: ce0368a

packages/protobuf21/PKGBUILD
cbix: 87c8feca3223f00ca4cdfad978f6904d8711c642

packages/protobuf21/PKGBUILD
cbix: 87c8feca3223f00ca4cdfad978f6904d8711c642

PR #355, opened 2024-02-04
paulxstretch: update patch checksum

cbix:


manu-cyber:
Looks like the patch has changed again? Right now I’m getting sha256sum '4dd7ce8eaba1f90c514c9689b63729db8ecc482f6fba087cef9a960792ce3a3b' for the patch file. Can you tell whether the current patch version is legit?

PR #398, opened 2024-03-31
ninjam-server: use official git source

cbix:
- updates will need to bump the `_commit`
- Not sure if it makes sense at all to include a `pkgver()`


packages/ninjam-server/PKGBUILD
```diff
@@ -5,37 +5,44 @@
...
-sha256sums=('64a5e2745ed335cc3e750dafde4f5b80405762cbda4def534486ea081adf3eac'
+sha256sums=('4f78057d1f49eb192daed33d2a9c856bf1c56326b2716d35656955bb99450470'
             '3b88687ecfc7ec413bc530ae43a7c13f0924a9a2533c05286903e86bf46dafa0'
             '78a06cd88f3b721237f81436fc8653b5dc6d29ce25eb4ccf1179a51b77b9b4cc')
 backup=("etc/$_name.cfg")
 
+pkgver() {
+  cd $_name/$_name/server
+  sed -nr 's/#define VERSION "v(.+)"$/\1/p' ninjamsrv.cpp
+}
```
SpotlightKid, line 29: I don't think we should have a dynamic version in a non-VCS package.

If you use Git sources for a release package, define a tag or commit to use and set the version with the `pkgver` variable as usual.

packages/ninjam-server/PKGBUILD
```diff
@@ -5,37 +5,44 @@
...
 depends=(gcc-libs)
+makedepends=(git)
 groups=(pro-audio)
-conflicts=(ninjam-server-git)
-source=("$pkgname-$pkgver.tar.gz::https://github.com/cbix/$_name/archive/refs/tags/ninjamsrv/v$pkgver.tar.gz"
+source=("git+https://www-dev.cockos.com/ninjam/ninjam.git#commit=$_commit"
         "$_name.service"
         "$_name.sysusers")
-sha256sums=('64a5e2745ed335cc3e750dafde4f5b80405762cbda4def534486ea081adf3eac'
+sha256sums=('4f78057d1f49eb192daed33d2a9c856bf1c56326b2716d35656955bb99450470'
```
SpotlightKid, line 21: Using a checksum with a Git source requires pacman/makepkg >= 6.1, which currently Manjaro does not have yet.

Since we sync this package to the AUR, maybe we should wait with this change, until pacman 6.1 is available everywhere?

SpotlightKid (changes requested):

packages/ninjam-server/PKGBUILD
cbix: Yea, I agree, just checked the official Arch Linux packages and there is no package using `pkgver()` 👍 9c90ca3b4d0e4f337def9e3dd679b5c0176e03b6

packages/ninjam-server/PKGBUILD
cbix: Good point, just checked and it's failing on Manjaro. There's no hurry to merge this anyway, so let's just wait until it's supported :)

cbix: ping @SpotlightKid

SpotlightKid (approved):

PR #399, opened 2024-04-01
Package midi2-workbench

cbix:
- [MIDI 2.0 Workbench](https://github.com/midi2-dev/MIDI2.0Workbench)
- [Electron package guidelines](https://wiki.archlinux.org/title/Electron_package_guidelines)
- TODO
  - [x] fix repro
  - [x] fix namcap `Cross-directory hardlink in package`
  - [x] don't use prebuilt binaries from usb package

### DIFF
--- a/nvchecker/archlinux-proaudio.toml
+++ b/nvchecker/archlinux-proaudio.toml
@@ -235,6 +235,13 @@ use_max_tag = true
 include_regex = '^v\d+(\.\d+)+$'
 prefix = "v"
 
+[midi2-workbench]
+source = "github"
+github = "midi2-dev/MIDI2.0Workbench"
+use_max_tag = true
+from_pattern = '^v(.*)p$'
+to_pattern = '\1'
+
 [midimonster]
 source = "github"
 github = "cbdevnet/midimonster"
@@ -489,14 +496,14 @@ github = "mhampton/ZetaCarinaeModules"
 use_max_tag = true
 prefix = "v"
 
-[ykchorus]
-source = "github"
-github = "SpotlightKid/ykchorus"
-use_max_tag = true
-prefix = "v"
-
 [xtuner]
 source = "github"
 github = "brummer10/XTuner"
+use_max_tag = true
+prefix = "v"
+
+[ykchorus]
+source = "github"
+github = "SpotlightKid/ykchorus"
 use_max_tag = true
 prefix = "v"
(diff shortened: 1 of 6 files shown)


nvchecker/archlinux-proaudio.toml
```diff
@@ -489,14 +496,14 @@ github = "mhampton/ZetaCarinaeModules"
...
 use_max_tag = true
 prefix = "v"
 
-[xtuner]
+[ykchorus]
 source = "github"
-github = "brummer10/XTuner"
+github = "SpotlightKid/ykchorus"
 use_max_tag = true
 prefix = "v"
```
cbix, line 509: these were just in the wrong order, ignore

packages/midi2-workbench/PKGBUILD
```diff
@@ -0,0 +1,66 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+_name=MIDI2.0Workbench
+pkgname=midi2-workbench
+pkgver=1.5.6
+_ver=${pkgver}p
+pkgrel=1
+pkgdesc='MIDI 2.0 debugging utility'
+arch=(x86_64)
```
cbix, line 10: Arch Linux ARM doesn't have electron and I personally wouldn't bundle it only for that reason.

packages/midi2-workbench/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+  ln -sf /usr/bin/node-gyp node_modules/.bin/node-gyp
+}
+
+build() {
+  cd $_name-$_ver
+  npx "${_npmargs[@]}" electron-builder -- \
+    --linux --$_arch --dir \
+    -c.electronDist=/usr/lib/$_electron \
+    -c.electronVersion=$_electron_ver \
+    -c.asar=false \
```
cbix, line 49: electron-builder giving a warning _"asar usage is disabled — this is strongly not recommended" but IMO there are more pros than cons to not creating a bundle
pros:
- we get a proper debug package with debug symbols
- package gets inspectable by namcap
- we can modify package contents in `package()`
- more transparent what gets installed
cons:
- larger installation size (~ 50 MiB vs 13 MiB)

cbix:
> using electron28 gives build errors

hm, that's surprising, since the upstream repo even still uses v21 🤔 Maybe the error is related to something else?

cbix:
electron27 seems to work, going with that for now until 29 turns stable :)

PR #427, opened 2024-06-04
Package fretboard

cbix:
- [Fretboard](https://apps.gnome.org/Fretboard/)
- TODO
  - [x] check license
  - [ ] fix reproducible build


packages/fretboard/PKGBUILD
```diff
@@ -0,0 +1,31 @@
...
+groups=(pro-audio)
+depends=(gtk4)
+makedepends=(blueprint-compiler git libadwaita meson rust)
+source=("git+https://github.com/bragefuglseth/$pkgname#tag=v$pkgver-fixed")
+sha256sums=('1125cbfc343c05a1cd2ca91c6cbeda6cb2e3f237e77ee841f500fe8c252b65e4')
+
+build() {
+  export RUSTUP_TOOLCHAIN=stable
+  # https://github.com/bragefuglseth/fretboard/issues/51
+  arch-meson --reconfigure -Dsandboxed=true -Dprofile=development $pkgname build
```
cbix, line 20: Here I'm configuring a debug build and relying on makepkg's binary optimizations just so the build is reproducible. This increases the package + install size by ~300 kB and ~1,3 MB, respectively but that's fine IMO until we find a way to make `cargo --release` output reproducible.

packages/fretboard/PKGBUILD
```diff
@@ -0,0 +1,31 @@
...
+# Contributor: Florian Hülsmann <fh@cbix.de>
+
+pkgname=fretboard
+pkgver=7.0
+pkgrel=1
+pkgdesc='Simple guitar chord lookup tool'
+arch=(aarch64 x86_64)
+url='https://apps.gnome.org/Fretboard/'
+license=(GPL-3.0-or-later)
+groups=(pro-audio)
```
cbix, line 11: Not sure if it makes sense to add `gnome-extra` since it's featured on https://apps.gnome.org/

packages/fretboard/PKGBUILD
cbix: Nvm, it just randomly worked for 1 run

cbix:
This is [now in extra](https://archlinux.org/packages/extra/x86_64/fretboard/)

jujudusud:
Nice! Thanks for the info! :-)

PR #435, opened 2024-06-13
feat: add new package 'stomptuner'

SpotlightKid:


packages/stomptuner/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+  'vst3-host: for loading the VST3 format plugins'
+)
+groups=(clap-plugins lv2-plugins pro-audio vst-plugins vst3-plugins)
+_dpf_commit='51edc8b6e3c636f29b5fca801bfb2551b0634b91'
+_pugl_commit='e33b2f6b0cea6d6263990aa9abe6a69fdfba5973'
+source=(
+  "${pkgname}-$pkgver.tar.gz::https://github.com/brummer10/$_name/archive/refs/tags/v$pkgver.tar.gz"
+  "DPF-$_dpf_commit.tar.gz::https://github.com/DISTRHO/DPF/archive/$_dpf_commit.tar.gz"
+  "pugl-$_pugl_commit.tar.gz::https://github.com/DISTRHO/pugl/archive/$_pugl_commit.tar.gz"
+)
```
cbix, line 28: IMO we should just use a git source with submodules, see for example the `vcvrack` PKGBUILD or [Arch Wiki](https://wiki.archlinux.org/title/VCS_package_guidelines#Git_submodules). The stomptuner repo could then have `#tag=v$pkgver` and a sha256sum for verification, which is then not necessary for the submodule sources. This would make it easier to update in the future :)

cbix (changes requested):

packages/stomptuner/PKGBUILD
SpotlightKid: Done.

cbix: just pushed a small fix, hope that's ok 🙏

cbix (approved):

packages/stomptuner/PKGBUILD
SpotlightKid: Yeah, good that you checked, I had a checksum there and then ran `updpkgsums` after I changed the Github URL and didn't check the `PKGBUILD` afterwards again. I thought that it now handles checksums for git repos, but apparently it doesn't.

cbix: It does, just seems the extra `?` confused it :D

PR #439, opened 2024-06-17
feat: add new package 'cloudreverb'

SpotlightKid:


packages/cloudreverb/PKGBUILD
```diff
@@ -0,0 +1,73 @@
...
+  'jack: for running the standalone version with JACKs'
+  'lv2-host: for loading the LV2 plugin'
+  'vst3-host: for loading the VST3 plugin'
+)
+groups=(lv2-plugins pro-audio vst3-plugins)
+_juce_commit='b08520c2de1771af3dfcbfbc0e0b6b0b5eb083b0'
+source=(
+  "$pkgname-$pkgver.tar.gz::https://github.com/xunil-cloud/$_name/archive/refs/tags/v$pkgver.tar.gz"
+  "JUCE-$_juce_commit.tar.gz::https://github.com/juce-framework/JUCE/archive/$_juce_commit.tar.gz"
+)
```
SpotlightKid, line 26: I didn't use a `git` source & a JUCE submodule here, because checking out the whole JUCE repo is inefficient.

packages/cloudreverb/PKGBUILD
```diff
@@ -0,0 +1,73 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
+
+_name=CloudReverb
+pkgname=${_name,,}
+_plugin_uri='http://github.com/xunil-cloud/$_name'
+pkgver=0.2
+pkgrel=1
+pkgdesc='An algorithmic reverb plugin/app for creating huge, endless spaces and modulated echoes'
```
cbix, line 9: Not a blocker, but in vim the `/` causes this to be marked faulty due to [PKGBUILD.vim](https://gitlab.archlinux.org/pacman/pacman-contrib/-/blob/master/src/vim/syntax/PKGBUILD.vim?ref_type=heads#L47) from pacman-contrib.

packages/cloudreverb/PKGBUILD
```diff
@@ -0,0 +1,73 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
+
+_name=CloudReverb
+pkgname=${_name,,}
+_plugin_uri='http://github.com/xunil-cloud/$_name'
```
cbix, line 6: Should be
```suggestion
_plugin_uri="http://github.com/xunil-cloud/$_name"
```
in order to resolve the variable in this place, right?

packages/cloudreverb/PKGBUILD
```diff
@@ -0,0 +1,73 @@
...
+pkgrel=1
+pkgdesc='An algorithmic reverb plugin/app for creating huge, endless spaces and modulated echoes'
+arch=(x86_64 aarch64)
+url='https://github.com/xunil-cloud/CloudReverb'
+license=(MIT)
+depends=(alsa-lib freetype2 glibc gcc-libs)
+makedepends=(cmake webkit2gtk)
+checkdepends=(lv2lint)
+optdepends=(
+  'jack: for running the standalone version with JACKs'
```
cbix, line 17: I only ever use one JACK at a time :P
```suggestion
  'jack: for running the standalone version with JACK'
```

packages/cloudreverb/PKGBUILD
cbix: Another idea would be to devendor JUCE, as done in some upstream packages, for example
- [ob-xd](https://gitlab.archlinux.org/archlinux/packaging/packages/ob-xd/-/blob/main/PKGBUILD)
- [elephantdsp-roomreverb](https://gitlab.archlinux.org/archlinux/packaging/packages/elephantdsp-roomreverb/-/blob/main/PKGBUILD)
- [surge-xt](https://gitlab.archlinux.org/archlinux/packaging/packages/surge-xt/-/blob/main/PKGBUILD)

Could be challenging though, some project depend too much on juce being a submodule or a specific version etc.

cbix (changes requested):
Left some minor remarks. Should't this be almost identical to Aether? 😅

packages/cloudreverb/PKGBUILD
SpotlightKid: Already tried that. Didn't work.

cbix (approved):

Issue #448, opened 2024-07-15
ola doesn't build with gcc 14.1.1


cbix:
- See OpenLightingProject/ola#1966

topas-rec:
For me ola currently doesn't build with the current PKGBUILD from [AUR ola](https://aur.archlinux.org/packages/ola) on arch.
I guess, this issue is about this problem, since I have the same problems as the linked issue.

I have a patch that fixes the build and creates a package that c
an be used successfully.
[fix_build_with_gcc_14.zip](https://github.com/user-attachments/files/16729632/fix_build_with_gcc_14.zip)

I wonder how to proceed, because:

* the patch only works for the source code from the release [0.10.9](https://github.com/OpenLightingProject/ola/releases/tag/0.10.9) of ola
* ola master branch needs slightly different fixes.

Because of this I think doesn't make sense to submit this patch to upstream.
My suggestion is that
* this patch is added to the arch PKGBUILD (the package)
* I try to create a pull request that fixes the issue upstream

Do you agree?

topas-rec:
<details>PKGBUILD that uses the patch

<summary>PKGBUILD that uses the patch</summary>

```
# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
# Contributor: Florian Hülsmann <fh@cbix.de>

pkgbase=ola
pkgname=(ola ola-docs)
pkgver=0.10.9
pkgrel=7
pkgdesc='Open Lighting Architecture for controlling entertainment lighting equipment'
arch=(x86_64 aarch64)
url='https://www.openlighting.org'
license=(LGPL2.1 GPL2)
depends=()
makedepends=(avahi cppunit doxygen flake8 libftdi-compat liblo libmicrohttpd ncurses protobuf-21
  python-numpy python-protobuf util-linux-libs)
source=("$pkgbase-$pkgver.tar.gz::https://github.com/OpenLightingProject/$pkgbase/archive/refs/tags/$pkgver.tar.gz"
        "fix-protobuf-v20.patch::https://github.com/OpenLightingProject/$pkgbase/pull/1875.patch"
        "fix-liblo.patch::https://github.com/OpenLightingProject/$pkgbase/pull/1954.patch"
        "fix_build_with_gcc_14.patch")
sha256sums=('d2a80d907f5a0a5169583060bd32aa0fdbb2fdfe18d1194dd922595ab6d447b5'
[...]

PR #463, opened 2024-08-23
fix: add patch to fix building ola with GCC 14

SpotlightKid:
* Also update license identifiers

Fixes: #448

### DIFF
--- a/packages/ola/PKGBUILD
+++ b/packages/ola/PKGBUILD
@@ -4,28 +4,31 @@
 pkgbase=ola
 pkgname=(ola ola-docs)
 pkgver=0.10.9
-pkgrel=7
+pkgrel=8
 pkgdesc='Open Lighting Architecture for controlling entertainment lighting equipment'
 arch=(x86_64 aarch64)
 url='https://www.openlighting.org'
-license=(LGPL2.1 GPL2)
+license=(GPL-2.0-or-later LGPL-2.1-or-later)
 depends=()
 makedepends=(avahi cppunit doxygen flake8 libftdi-compat liblo libmicrohttpd ncurses protobuf-21
   python-numpy python-protobuf util-linux-libs)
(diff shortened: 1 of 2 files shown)


SpotlightKid:
> Seems to fail on aarch64 🤔

Yeah, but why?

```
error: Could not create output directory /tmp/build/ola/src/ola-0.10.9/html
make: *** [Makefile:19203: doxygen-doc/ola.tag] Error 1
==> ERROR: A failure occurred in build().
    Aborting...
```

Could this be a disk space issue?

cbix:
Ah, could be, just ran a `docker system prune` and restarted, thanks 👍

SpotlightKid:
Now we have a test failure on aarch64. Looks like a timing / performance issue.

Should we just exclude the test? There's a dozen test in this one test method, though.

```
...F..

======================================================================

FAIL: testEventLoop (__main__.ClientWrapperTest.testEventLoop)
----------------------------------------------------------------------

  File "/tmp/build/ola/src/ola-0.10.9/./python/ola/ClientWrapperTest.py", line 209, in testEventLoop
    self.assertAlmostEqual(d_diff, datetime.timedelta(milliseconds=15),
AssertionError: datetime.timedelta(microseconds=16697) != datetime.timedelta(microseconds=15000) within datetime.timedelta(microseconds=750) delta (datetime.timedelta(microseconds=1697) difference)

----------------------------------------------------------------------
Ran 6 tests in 0.019s

FAILED (failures=1)
FAIL python/ola/ClientWrapperTest.sh (exit status: 1)
```

SpotlightKid:
Could not reproduce the test failure in two further CI runs. The notification step still doesn't work correctly. Not sure why. Need to re-check the authentication token.

Other than that, I think this can be merged.

cbix (approved):

PR #468, opened 2024-09-04
Revert "hack: install previous fakeroot version and ignore future updates"

cbix:
Reverts osam-cologne/archlinux-proaudio#434

- Issue seems to be related to the CI host, I'll try upgrading to a newer ubuntu LTS


cbix:
Apparently upgrading the CI host to ubuntu 22.04 fixed it 😂

PR #474, opened 2024-09-23
jamulus: update to v3.11.0

cbix:


SpotlightKid:
Package `qt6-tools`, which provides `/usr/lib/qt6/bin/lrelease`, needs to be added to makedepends.

Issue #445, opened 2024-07-08
Package jack_link


tkna91:
Make sure the software
- [x] does not already exist in the official Arch Linux repos: https://archlinux.org/packages/
- [x] does not already exist in this repo: https://arch.osamc.de/#packages
- [x] is released under a free software license
- [x] has released a stable version (otherwise ask the developer to tag a release!)

- [x] (if possible) can be successfully built on a x86_64/amd64 machine

- Name: `jack_link`
- Suggest a [package name](https://wiki.archlinux.org/title/PKGBUILD#pkgname): `jack_link`
- Short [description](https://wiki.archlinux.org/title/PKGBUILD#pkgdesc): `JACK transport timebase bridge to Ableton Link`
- Latest release version: `0.2.2`
- Website: https://github.com/rncbc/jack_link
- Source code: https://github.com/rncbc/jack_link.git
- AUR (if exists): `none`
- License(s): `GNU General Public License, version 2`
- Is the project active? `yes`
- I have successfully built this myself on Arch: `yes`

This software is from the same author who makes qtrator, qjackctl, qpwgraph, etc. Please consider packaging this software.

PR #524, opened 2025-01-25
feat: add new package 'jack_link'

SpotlightKid:
Implements: #445

### DIFF
--- a/packages/jack_link/PKGBUILD
+++ b/packages/jack_link/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
+
+_link_commit="41d9aa111f702e78b6fbaee9d3e06dda1db6420d"
+pkgname=jack_link
+pkgver=0.2.3
+pkgrel=1
+pkgdesc='JACK transport bridge to Ableton Link'
+arch=(aarch64 x86_64)
+url='https://github.com/rncbc/jack_link'
+license=(GPL-2.0-or-later)
+depends=(gcc-libs glibc)
+makedepends=(asio jack)
+groups=(pro-audio)
+source=("$pkgname-$pkgver.tar.gz::https://github.com/rncbc/jack_link/archive/refs/tags/v$pkgver.tar.gz"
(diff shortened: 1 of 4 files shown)


packages/jack_link/PKGBUILD
```diff
@@ -0,0 +1,36 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
+
+_link_commit="41d9aa111f702e78b6fbaee9d3e06dda1db6420d"
+pkgname=jack_link
```
cbix, line 5: My first thought was that I've never seen a pkgname with an underscore, but apparently that's perfectly fine according to the wiki, and `jack_capture` is another example doing that 👍

packages/jack_link/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+pkgrel=1
+pkgdesc='JACK transport bridge to Ableton Link'
+arch=(aarch64 x86_64)
+url='https://github.com/rncbc/jack_link'
+license=(GPL-2.0-or-later)
+depends=(gcc-libs glibc)
+makedepends=(asio jack)
+groups=(pro-audio)
+source=("$pkgname-$pkgver.tar.gz::https://github.com/rncbc/jack_link/archive/refs/tags/v$pkgver.tar.gz"
+        "link-$_link_commit.tar.gz::https://github.com/Ableton/link/archive/$_link_commit.tar.gz")
```
cbix, line 16: I personally found that since pacman supports checksums for VCS sources it's slightly easier to maintain packages using a git source and the submodule head being checked out at the correct commit automatically, see e.g. stomptuner or vcvrack.

cbix (approved):
Nice work 🎉 Just nits, no blockers!

packages/jack_link/PKGBUILD
SpotlightKid: Even the official repos have packages with underscores in their names.

packages/jack_link/PKGBUILD
SpotlightKid: That then also requires the main project sources to be a git URL, because you then need the `.gitmodules` file.

I like the state of sub-module projects to be a little more explicit in the PKGBUILD.

PR #526, opened 2025-01-29
Basic PKGBUILD templates

cbix:


SpotlightKid (commented):
These files look good to me, but I'm wondering how useful they will be in their minimal form.

Wouldn't it be better to have files with example data, where you you can just use search & replace when creating a new package?

Usually I don't need help remembering the syntax and field names of PKGBUILD files but I need to look up the common values for groups, optdepends, etc. and the shell idioms and command options commonly used so that I can copy & paste them into the new PKGBUILD.

IMO it would be easier to remove the stuff you don't need from a template than having to remember the stuff you need to add in.

cbix:
@SpotlightKid good point, you mean something like 1b0980ccd54653c10c2b1e270160678bf1ca73b7?

SpotlightKid:
> @SpotlightKid good point, you mean something like [1b0980c](https://github.com/osam-cologne/archlinux-proaudio/commit/1b0980ccd54653c10c2b1e270160678bf1ca73b7)?

Yes, exactly.

cbix:
@SpotlightKid perfect, there you go: fda5b53704d0b5509b2595f0dfeabb65035b53ff

SpotlightKid (approved):

PR #528, opened 2025-01-30
Split up nvchecker config

### LINKED ISSUE #275: Dynamically build nvchecker old_ver from package sources
Currently we have to manually keep the `nvchecker/old_ver.json` up to date with the current package versions. That step could be eliminated by generating that json from the package sources in `run-nvchecker.cron`. This comes with two prerequisites:
1. package names have to exactly match the names in `archlinux-proaudio.toml` ✅
2. nvchecker version formats have to match the `pkgver` values ✅

cbix:
- Closes #275 
- TODO:
  - [x] update README
  - after merging, manually pull the latest master branch on the server, otherwise the executed cron script won't generate the config on its next run


nvchecker/README.md
```diff
@@ -90,7 +90,6 @@ Command line options passed to `run-nvchecker.cron` are passed on to
 
 Note: by default, the script reporrts new version only once and records the
```
SpotlightKid, line 91: "reporrts" -> "reports"

nvchecker/gen-config.sh
```diff
@@ -0,0 +1,25 @@
...
+
+__EOF__
+
+VER_TEMP="$(mktemp)"
+for f in packages/*/.nvchecker.toml; do
+  cat $f >> $CFG
+  echo "" >> $CFG
+  pkgdir=$(dirname $f)
+  pkgbase=$(basename $pkgdir)
+  pkgver=$(. $pkgdir/PKGBUILD; echo $pkgver)
```
SpotlightKid, line 21: I'm not sure how I feel about this. Since this is executed on the repo server directly, this seems like an avenue too easy to exploit for executing arbitrary stuff. Aren't there any tools, which can parse the information from PKGBUILD files more safely?

packages/jack_link/.nvchecker.toml
```diff
@@ -1,5 +1,5 @@
 [jack_link]
-source = "git"
+source = "github"
```
SpotlightKid, line 2: Thanks for catching those! My bad.

nvchecker/gen-config.sh
```diff
@@ -0,0 +1,24 @@
...
+VER_TEMP="$(mktemp)"
+for f in packages/*/.nvchecker.toml; do
+  echo "" >> $CFG
+  cat $f >> $CFG
+  pkgdir=$(dirname $f)
+  pkgbase=$(basename $pkgdir)
+  pkgver=$(. $pkgdir/PKGBUILD; echo $pkgver)
+  printf '{"%s": "%s"}\n' $pkgbase $pkgver >> "$VER_TEMP"
+done
+jq -s add "$VER_TEMP" > $VER
```
SpotlightKid, line 24: Please add a note to the readme that `jq` needs to be installed to run the `run-nvchecker.cron` script.

nvchecker/run-nvchecker.cron
```diff
@@ -9,5 +9,6 @@ cd "$CHECKOUT"
 git stash
 git pull -q
 git stash pop
+nvchecker/gen-config.sh
```
SpotlightKid, line 12: We should catch errors here explicitly or make the whole script run with the `-e` option.

SpotlightKid (changes requested):
Please rebase and see comments.

nvchecker/gen-config.sh
cbix: Good point, my reasoning was basically that the server pulls only from `master` which we can (and should) expect to trust, otherwise the attack could as well just happen inside one of these nvchecker scripts itself.

There are probably 10 more ways to do the same thing, but one idea could be:
1. CI generates nvchecker configs on push to master
2. configs get uploaded to website
3. cronjob (either on the existing server or from a [CI cronjob](https://docs.drone.io/cron/) pulls configs from website to execute

regarding just parsing the PKGBUILD here, this should do:
```suggestion
  pkgver=$(sed -n 's/^pkgver=//p' $pkgdir/PKGBUILD)
```

SpotlightKid: That wouldn't handle quotes correctly, wouldn't it? They would become part of the value.

nvchecker/gen-config.sh
cbix: 46f20ae7ddeedb7f89c7390be9791f93cc1e6bc9

nvchecker/run-nvchecker.cron
cbix: 8074a9ce14cfaf79618a4f9ddc1e5c633e7ab407

nvchecker/gen-config.sh
cbix: 27e83b4f24bdcb8edb0bf47c1835ee638c1e1ea9

nvchecker/README.md
cbix: f66290a480a33eae925c3da40cd1a5b39ec8d7dd

nvchecker/gen-config.sh
cbix: fixed in 62fb3d9d1a0ee590946210f57523dc07a8499e8f

SpotlightKid: ok

SpotlightKid (approved):

PR #543, opened 2025-02-15
ratatouille.lv2: update to v0.9.10

### LINKED ISSUE #570: ratatouille.lv2 freezes during `check` step
The `check()` function in the ratatouille.lv2 PKGBUILD has an `lv2lint` command that seems to hang when complete, never returning control back to the calling process. This happened both when using paru and when cloning the AUR repo and running `makepkg -si` directly. I was able to complete the installation by commenting out the entire `check()` function and rerunning `makepkg -si`.

Thanks!

cbix:
- brummer10/Ratatouille.lv2#42
- brummer10/Ratatouille.lv2#45
- Closes #570


cbix:
closing for now, g++ is expected to be present so it's reasonable to explicitly build using it

PR #595, opened 2025-04-20
octasine: disable LTO to fix missing symbol

cbix:
- workaround for greatest-ape/OctaSine#203
- Still not sure how this issue got introduced, could be recent changes in makepkg or rust 🤔
- Disabling LTO might decrease performance
- testing CLAP can be done with `clap-example-host -p pkg/octasine/usr/lib/clap/OctaSine.clap`


packages/octasine/PKGBUILD
```diff
@@ -1,15 +1,16 @@
...
 # Contributor: Christopher Arndt <aur -at- chrisarndt -dot- de>
 # Contributor: Mathias Buhr <napcode@aparatus.de>
+# Contributor: Florian Hülsmann <fh@cbix.de>
 
 _name=OctaSine
 pkgname=${_name,,}
 pkgver=0.9.1
-pkgrel=1
+pkgrel=2
 pkgdesc='A four-operator stereo FM synthesizer VST2 plugin'
```
SpotlightKid, line 10: We should update the description to either add 'CLAP' or remove 'VST2'.

SpotlightKid (approved):
I can confirm that the plugins, both CLAP and VST2, now are detected and properly loaded again and function normally.

I ran an old project with +10 instances of OctaSine (VST2) in Ardour and could not decern a significant performance degradation.

packages/octasine/PKGBUILD
cbix: 9708cc8fa40ef3f0eb75faa6b038ec25d384d7b5

PR #605, opened 2025-05-03
Open nvchecker PRs from cron

cbix:
The goal is to have a CI pipeline automatically open PRs for new versions, which are then manually reviewed and merged. This should save us some manual effort and time.


cbix:
Let's see if this works well, otherwise it should be a simple revert or disabling the cron event in drone.

PR #619, opened 2025-06-01
Split package ratatouille

### LINKED ISSUE #563: Ratatouille split package
Hi
Just to let you know, Ratatouille.lv2 contains now a stand alone version, which build beside the plugin, by default when run plain make.
In case, you would prefer to have only the lv2 plugin in the package you should consider to run 
`make lv2`  to get the plain lv2 plugin
or `make modapp` to get the additional MOD interface to be usable with the mod desktop app.

The stand alone version alone, could be build separate with `make standalone` 

As this beehive has changed in the last release, I think I write you a note so you could decide how to handle that.

regards
and thanks for taking the time to distribute Ratatouille.
hermann

cbix:
- Closes #563


packages/ratatouille/PKGBUILD
```diff
@@ -0,0 +1,104 @@
...
+
+_pick() {
+  local p="$1" f d; shift
+  for f; do
+    d="$srcdir/$p/${f#$pkgdir/}"
+    mkdir -p "$(dirname "$d")"
+    mv "$f" "$d"
+    rmdir -p --ignore-fail-on-non-empty "$(dirname "$f")"
+  done
+}
```
cbix, line 30: this snippet is often used in official arch packages

packages/ratatouille/PKGBUILD
```diff
@@ -0,0 +1,104 @@
...
+  export CXXFLAGS="${CXXFLAGS/-Wp,-D_FORTIFY_SOURCE=3/}"
+  make CXX=g++ STRIP=:
+}
+
+check() {
+  cd $_name-v$pkgver
+  lv2lint -M pack -s '_Z*' -I bin/$_name "$_plugin_uri"
+}
+
+package_ratatouille() {
```
cbix, line 44: Not sure but does it make sense to set `arch=(any)` here since it's just a meta package?

packages/ratatouille/PKGBUILD
```diff
@@ -0,0 +1,102 @@
...
+  cd $_name-v$pkgver
+  make DESTDIR="$pkgdir" install
+  cd "$pkgdir"
+  _pick $pkgbase-clap usr/lib/clap/*
+  _pick $pkgbase-lv2 usr/lib/lv2/*
+  _pick $pkgbase-standalone usr/bin/*
+  _pick $pkgbase-standalone usr/share/{applications,pixmaps}
+  _pick $pkgbase-vst usr/lib/vst/*
+}
+
```
SpotlightKid, line 56: namcap also says:

    Uncommon license identifiers such as 'BSD-3-Clause' require license files below /usr/share/licenses/ratatouille/ or switching to common license identifiers. Found 0/1 required license files.

Not sure, whether this makes sense, give n that the package is empty.

A similar case in the extra repo is https://archlinux.org/packages/extra/x86_64/sfizz/ and it does have license file.

packages/ratatouille/PKGBUILD
```diff
@@ -0,0 +1,102 @@
...
+  cd $_name-v$pkgver
+  install -vDm644 LICENSE -t "$pkgdir"/usr/share/licenses/$pkgname
+}
+
+package_ratatouille-docs() {
+  arch=(any)
+  pkgdesc+=" (documentation)"
+  cd $_name-v$pkgver
+  install -vDm 644 README.md Ratatouille.png -t "$pkgdir"/usr/share/doc/$pkgbase
+}
```
SpotlightKid, line 72: ratatouille-docs E: Uncommon license identifiers such as 'BSD-3-Clause' require license files below /usr/share/licenses/ratatouille-docs/ or switching to common license identifiers. Found 0/1 required license files.

packages/ratatouille/PKGBUILD
cbix: b66742eb3727b02cd5d32e4ce332496dad63bc03

packages/ratatouille/PKGBUILD
```diff
@@ -0,0 +1,102 @@
...
+  optdepends=('lv2-host: for loading the LV2 plugin')
+  groups+=(lv2-plugins)
+  mv -v $pkgname/* "$pkgdir"
+  cd $_name-v$pkgver
+  install -vDm644 LICENSE -t "$pkgdir"/usr/share/licenses/$pkgname
+}
+
+package_ratatouille-standalone() {
+  pkgdesc+=" – JACK standalone"
+  depends+=(libcairo.so libjack.so libportaudio.so libsndfile.so)
```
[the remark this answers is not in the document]
cbix: When adding libasound as a dependency, namcap tells me it's an "unused dependency", which can be confirmed using `ldd -u`:
```
Unused direct dependencies:
	/usr/lib/libasound.so.2
```
i.e. it's in the ELF header but apparently not used.

packages/ratatouille/PKGBUILD
cbix: 0d48eab64d804640c905cc4338870cbd5d2fbe91

packages/ratatouille/PKGBUILD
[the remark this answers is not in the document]
SpotlightKid: Does that mean, that the program can start without `libasound.so` being present?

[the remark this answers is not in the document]
cbix: I quickly tested how this would behave with a small hello world linked to a random library which I could then remove, and indeed I'm getting an error in that case, so let's add it: fc4678683906bcec647eb254e89215e1f8adbf3f

SpotlightKid (approved):

PR #634, opened 2025-06-23
feat: add new package 'python-freesimplegui'

SpotlightKid:


packages/python-freesimplegui/PKGBUILD
```diff
@@ -0,0 +1,37 @@
+# Maintainer: OSAMC <https://github.com/osam-cologne/archlinux-proaudio>
+# Contributor: Christopher Arndt <aur at chrisarndt dot de>
+
+_projectname=FreeSimpleGUI
+_name=${_projectname,,}
+pkgname=python-$_name
+pkgver=5.2.0.post1
```
cbix, line 7: Apparently this is identical to 5.2.0, but do I understand it correctly that the added `pyproject.toml` is kind of important? This is just a weird version string to use :D

packages/python-freesimplegui/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+pkgrel=1
+pkgdesc='A free, easy-to-use Python GUI framework based on tkinter'
+url='https://github.com/spyoungtech/FreeSimpleGUI'
+arch=(any)
+license=(LGPL-3.0-or-later)
+depends=(python)
```
cbix, line 13: PySimpleGui had `tk` as a dependency, does that apply here as well?

cbix (changes requested):
Just a few questions :)

PR #709, opened 2025-11-18
feat: remove package triceratops

SpotlightKid:
Dependency gtkmm is not available in Arch repos (only AUR) anymore


SpotlightKid:
Note that there is a newer version of triceratops (currently 0.8.1) on its homepage (which our nvchecker rule didn't catch, btw!), but this is a completely new re-implementation, based on DPF and a) it has a new LV2 URI and b) in my tests the UI did not draw correctly in any host I tried (jalv, Carla, Ardour). Same issue with the stand-alone JACK app version.

PR #790, opened 2026-04-13
Fix compilation of the abcm2ps package.

HugoNikanor:
As presented, the code doesn't compile on (at least my) system. This adds some basic compilation flags which works for me.


cbix (changes requested):
Thanks for this! Make sure to increment `pkgrel` :)

SpotlightKid:
Closing in favour of #791

PR #843, opened 2026-08-01
[neuralrack] update to 0.4.0

SpotlightKid:
* Remove obsolete prepare function
* Remove portaudio dependency (windows only)


cbix:
@SpotlightKid the makefile now builds vst3 as well, this should be extracted into a separate package

PR #851, opened 2026-08-04
jack_transport_link: use vercmp instead of parse_version

cbix:
- nvchecker would detect older releases as more recent due to upstream's messed up versioning scheme, see #850


SpotlightKid (approved):
```py
>>> pyalpm.vercmp("1.3.0", "1.4.0-linkaudio.5")
-1
```

LGTM 👍

PR #863, opened 2026-08-14
feat: add new package toneshifteq

SpotlightKid:
https://github.com/brummer10/ToneShiftEQ


packages/toneshifteq/PKGBUILD
```diff
@@ -0,0 +1,130 @@
...
+  mv -v $pkgname/* "$pkgdir"
+  cd $_name-v$pkgver
+  install -vDm 644 README.md Manual.md $_name.png -t "$pkgdir"/usr/share/doc/$pkgname
+  install -vDm 644 LICENSE -t "$pkgdir"/usr/share/licenses/$pkgname
+}
+
+package_toneshifteq-lv2() {
+  pkgdesc+=" – LV2"
+  depends+=(glibc libcairo.so libfftw3.so libfftw3f.so libfreetype.so libgcc
+            libstdc++ libx11)
```
SpotlightKid, line 100: For some reason the LV2 version does not link to libsndfile.

packages/toneshifteq/.nvchecker.toml
```diff
@@ -0,0 +1,6 @@
+[toneshifteq]
+source = "github"
+github = "brummer10/ToneShiftEQ"
+use_max_tag = true
+include_regex = '^v\d+(\.\d+)+$'
```
cbix: The include_regex seems to not be necessary so far
```suggestion
```

packages/toneshifteq/PKGBUILD
```diff
@@ -0,0 +1,130 @@
...
+
+  for spec in ${lv2specs[@]}; do
+    ln -vsf /usr/lib/lv2/$spec.lv2 .lv2
+  done
+
+  echo "Checking $_name.lv2 with lv2lint ..."
+  ln -vsf "../bin/$_name.lv2" .lv2
+  # LV2 plugin should not export all symbols, but it does.
+  # See https://github.com/brummer10/ToneShiftEQ/issues/8
+  LV2_PATH="${PWD}/.lv2" lv2lint -s '*' "$_plugin_uri"
```
cbix: `LV2_PATH`can be concatenated, see for example

https://github.com/osam-cologne/archlinux-proaudio/blob/c5a772f82f133382d159973c10e1a0c049950aef/packages/swh-plugins.lv2/PKGBUILD#L43


```suggestion
  LV2_PATH="${PWD}/.lv2:/usr/lib/lv2" lv2lint -s '*' "$_plugin_uri"
```
This _should_ make the above `ln -vsf` obsolete :)

packages/toneshifteq/PKGBUILD
cbix: Maybe it `dlopen`s it, but that would be weird :D

cbix (changes requested):

packages/toneshifteq/PKGBUILD
SpotlightKid: The whole point of this linking is to isolate the `lv2lint` check from *anything* installed in `/usr/lib/lv2` on the system. This might not matter in our CI, but since we also push to the AUR, I think this is important and also makes testing package builds locally easier.

cbix: Hmm good point, I'll look at orting this to the swh package eventually :)

cbix (approved):

